PR #1, opened 2022-10-06
Initial work on the underlying data structure + Package Setup

dougbrn:
The first stab at an underlying data structure is a multi-indexed pandas dataframe, where the band+an integer index describe the indexing scheme, and the flux, flux error, and time are the included data columns.

Included a from_dict I/O routine, which may or may not be useful in the long term, but is useful in the very short term.

Properties were added to allow easy access to the data columns/band index, but the multi-indexing may require these to be a bit more complex (option for band specific properties?)

I'd propose an initial convention for package shorthand as: 
`import lsstseries as ls`

### DIFF
--- a/lsstseries/timeseries.py
+++ b/lsstseries/timeseries.py
@@ -1,0 +1,41 @@
+import pandas as pd
+import numpy as np
+
+class timeseries():
+    def __init__(self,data=None):
+        self.data = data
+    
+    #I/O
+    def from_dict(self,data_dict,band_label='band'):
+        """Build dataframe from a python dictionary"""
+        index = self._build_index(data_dict[band_label])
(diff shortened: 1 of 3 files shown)


lsstseries/timeseries.py
```diff
@@ -0,0 +1,41 @@
+import pandas as pd
+import numpy as np
+
+class timeseries():
+    def __init__(self,data=None):
+        self.data = data
```
nevencaplar, line 6: With the fullest understanding this is only a prototype, so feel free to ignore; This allows the user to pass in the data that is not sorted, i.e., one can pass data that is not ready to be used in any way.

nevencaplar (approved):
Looks good!

PR #2, opened 2022-10-18
Ensemble Initial Implementation

dougbrn:
This PR includes a *rough* implementation of the ensemble class, which is the object that contains many lightcurves. This is distinct from the timeseries class, which holds the information of one lightcurve, indexed by band and an arbitrary integer index. The ensemble is essentially just one additional index level, based on an object id, which allows index slicing to return individual light curves.

The implementation here focuses on IO and particularly the use case of pulling data down from the TAP: https://data.lsst.cloud/api-aspect . A token is needed to query against TAP, the link explains how to generate one.

As I mention above, I think this implementation is somewhat rough. The initial challenge with trying to emulate a LightKurve like approach for lsstseries is that with TAP we are querying for multiple objects, rather than pulling down a single object via ID (a use case we will want to implement soon). The ADQL query space is huge, and so there's a trade-off between giving users the power to construct complex queries, and user friendliness.  I've attached a diagram of how the ensemble workflow currently functions for TAP. Note that realistically a user may query TAP several times in succession to get a final target list. The query_ids function exists to give a more constrained query that is formatted such that lightcurves can be pulled out of it.

This PR also builds out timeseries slightly, adding some handling for different column names and stored metadata.

ensemble drawio (2)

### DIFF
--- a/lsstseries/ensemble.py
+++ b/lsstseries/ensemble.py
@@ -87,12 +87,22 @@ class ensemble():
         cols = [time_col, flux_col, err_col]
 
         if to_mag:
-            query_cols, cols = self.flux_to_mag(cols)
+            flux_query, flux_label = self.flux_to_mag([flux_col])
+            flux_col = flux_label[0]
+            if err_col is not None:
+                err_query, err_label = self.flux_to_mag([err_col])
+                err_col = err_label[0]
+
+            query_cols = [time_col] + flux_query + err_query
+            cols = [time_col, flux_col, err_col]
+
         else:
             query_cols = cols
 
         if add_cols is not None:
             cols = cols+add_cols
+            query_cols = query_cols+add_cols
+
         idx_cols = ['diaObjectId', 'filterName']
 
         result = pd.DataFrame(columns=idx_cols+cols)
--- a/lsstseries/ensemble.py
+++ b/lsstseries/ensemble.py
@@ -42,8 +42,7 @@ class ensemble():
         cred = vo.auth.CredentialStore()
         cred.set_password("x-oauth-basic", self.token)
 
-        service = vo.dal.TAPService("https://data.lsst.cloud/api/tap", 
-                                    cred.get("ivo://ivoa.net/sso#BasicAA"))
+        service = vo.dal.TAPService("https://data.lsst.cloud/api/tap", cred.get("ivo://ivoa.net/sso#BasicAA"))
         results = service.search(query, maxrec=maxrec)
         result = results.to_table().to_pandas()
         self.result = result
@@ -53,10 +52,11 @@ class ensemble():
                   time_col='midPointTai',
                   flux_col='psFlux',
                   err_col='psFluxErr',
-                  add_cols=[],
+                  add_cols=None,
                   id_field='diaObjectId',
                   catalog='dp02_dc2_catalogs',
                   table='DiaSource',
+                  to_mag=True,
                   maxrec=None):
         """Query based on a list of object ids; applicable for DP0.2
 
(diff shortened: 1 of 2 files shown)


lsstseries/timeseries.py
```diff
@@ -2,8 +2,10 @@
 import numpy as np
 
 class timeseries():
-    def __init__(self,data=None):
-        self.data = data
+    def __init__(self):
+        self.data = None
+        self.id = None
+        self.cols = ['time','flux','flux_err']
```
drewoldag, line 8: Similarly, if `self.cols` is constant, then it could be moved to a class attribute I believe. And the references to `self.cols[i]` on lines 28, 33, 38, could be defined as class attributes as well. 

```
class timeseries():
    self._cols = ['time', 'flux', 'flux_err']
    self._time_col = self._cols.index('time')
```

But now that I look at line 20, maybe what I'm talking about isn't applicable, if `self.cols` can be redefined. If that's the case, do you have confidence that self.cols[0] will always be "time"?

dougbrn: Thanks for the good comments Drew, you're honing in on one of the things that I think needs improvement, regarding the relationship between the columns extracted from a given table in the ensemble object, and the level of enforcement on columns and their names in the timeseries class. I wouldn't have confidence that self.cols[0] will always be time, there should probably be some ability to map columns to the timeseries properties, with some defaults that work in most cases (I do think in most cases time will be the 0th column, and we have a level of influence on that in the design here). 

self.cols is currently overwritten by whatever the ensemble object passes in, and the index based property assignment (e.g. cols[0] is time) is a rough solution for it. What I think may make more sense would be to anticipate which columns from the TAP tables are types of fluxes, errors, and times. So that as those same column names are found in the ensemble objects, the timeseries object can appropriately allocate them to the right properties. I think in doing this the changes you suggest would be implementable

drewoldag: Oh, ok, I think I understand what you're getting at here. So, for instance, in ensemble.query_ids, the columns and order are defined for the query, and so you can have some confidence in the order of the columns being passed in by the ensemble. Is my understanding correct?

dougbrn: Exactly, and as you'd imagine right now it would be very easy for someone to alter the columns and make it incompatible with the timeseries class, so in general I think a lot more needs to exist that controls/infers the columns being passed between the two classes

jeremykubica: I'm not following all of this discussion. Shouldn't there be a few required columns that every time series has in order to be valid (time and flux)? If so, why would it cause a problem to always map those into known columns for our internal representation?

lsstseries/ensemble.py
```diff
@@ -0,0 +1,70 @@
+import pandas as pd
+import numpy as np
+import pyvo as vo
+from .timeseries import timeseries
+
```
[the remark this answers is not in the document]
nevencaplar: I see now that Jeremy has already made a similar comment

lsstseries/timeseries.py
dougbrn: Catching up on this, I think an enforced time and flux column is necessary, but there's two challenges that come with it:

1. Specifically pulling from the DP02 schema for the `DiaSource` table (https://dm.lsst.org/sdm_schemas/browser/dp02.html#DiaSource), there are multiple possible flux metrics available. E.g. a user may want `psFlux` or `apFlux` and the associated error quantities from them. From that, the mapping would either need to be able to infer that a given column is a "flux" column given some scheme (even as simple as having an internal list of accepted flux columns), or we could have a more explicit flux kwarg where the user specifies which column is the flux column to use in further analysis. Perhaps a combination of the two where the provided kwarg value is checked against a list of typical flux metrics, and a warning is thrown if it's not found?
2. The `query_tap()` function is extremely open-ended at the moment, as mentioned above it essentially just is a wrapper around an ADQL (Astronomy Data Query Language, SQL-like) interface. It can be used to query any of the DP02 tables (beyond `DiaSource`), and stores that as an internal query result. The challenge being that `query_tap()` can return tables that have no flux or time information. It makes sense to have `to_timeseries()` require properties from a given query result, but `query_tap()` gives query results that will be rejected by `to_timeseries()` in most cases (unless the user emulates the `query_ids()` function) which seems frustrating/unintuitive. As I'm typing this it doesn't really come off as an issue with the required columns as much as an issue with how `query_tap()`, which is ultimately a database exploratory tool, fits into the package.

Hope that sheds a bit more light!

lsstseries/ensemble.py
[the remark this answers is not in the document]
dougbrn: thanks for making these changes! I'll make sure future code conforms with this. I had opted to start with the short docstrings while the functions were potentially volatile (subject to major changes to functionality and arguments), but recognize that we should have the proper numpy-like docstrings in the PR code 🙏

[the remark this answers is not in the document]
nevencaplar: Np. It was also useful for me to familiarize with the code.

nevencaplar:
Second review -> Looks good to me

PR #7, opened 2022-10-29
Stetson j

nevencaplar:
Added function StetsonJ that compute StetsonJ variability index given a measurement of flux and a measurement of flux errors. It is possible to specify on which band to compute the index; if band is not specified the measurement is done on all avaliable bands.

### DIFF
--- a/lsstseries/timeseries.py
+++ b/lsstseries/timeseries.py
@@ -17,15 +17,16 @@ class timeseries():
         Parameters
         ----------
         data_dict : `dict`
-            Dictionary contaning the data
+            Dictionary contaning the data.
         time_label: `str`
-            Name for column containing time information
+            Name for column containing time information.
         flux_label: `str`
-            Name for column containing signal (flux, magnitude, etc) information
+            Name for column containing signal
+            (flux, magnitude, etc) information.
         err_label: `str`
-            Name for column containing error information
+            Name for column containing error information.
         band_label: `str`
-            Name for column containing filter information
+            Name for column containing filter information.
         """
 
         try:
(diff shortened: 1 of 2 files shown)


dougbrn (commented):
This looks good overall, two things:

1. Is it worth putting special handling for the case where you have 0 or 1 data points? In particular, should they return a NaN? I tried it with a dummy example and ran into that for the N=1 case, because it attempts an N/N-1 calculation. Just thinking about the use case where we have O(10^6+) running and don't want these to be erroring out necessarily
2. stetson_j_single has required parameters of fluxes and errors, meaning that a user would need to filter down to those band specific values manually for ingest into the function, would a single argument for band be better? This aligns it with stetson_J, but also would potentially make it redundant as it looks like stetson_J can handle a single input band string

lsstseries/timeseries.py
```diff
@@ -94,3 +96,141 @@ def _build_index(self, band):
...
+
+        Notes
+        ----------
+        In case that no value for band is passed, the function is executed
+        on all avaliable bands.
+        """
+
+        if band is None:
+            unq_band = np.unique(self.band)
+        if type(band) == 'str':
```
dougbrn, line 121: This should be str, not 'str', doesn't work as intended. Also, should handle case where user provides a subset of bands (e.g. band=['g','z']), as currently that would never assign unq_band and it would error out

nevencaplar: I have modified the code so that it supports passing a subset of bands - including the possibility to pass bands that do not exist in the data. I have incorporate that with if..else.. statment; not sure if this is long-term feasible.

> This looks good overall, two things:
> 
> 1. Is it worth putting special handling for the case where you have 0 or 1 data points? In particular, should they return a NaN? I tried it with a dummy example and ran into that for the N=1 case, because it attempts an N/N-1 calculation. Just thinking about the use case where we have O(10^6+) running and don't want these to be erroring out necessarily
> 2. stetson_j_single has required parameters of fluxes and errors, meaning that a user would need to filter down to those band specific values manually for ingest into the function, would a single argument for band be better? This aligns it with stetson_J, but also would potentially make it redundant as it looks like stetson_J can handle a single input band string

nevencaplar:
> This looks good overall, two things:
> 
> 1. Is it worth putting special handling for the case where you have 0 or 1 data points? In particular, should they return a NaN? I tried it with a dummy example and ran into that for the N=1 case, because it attempts an N/N-1 calculation. Just thinking about the use case where we have O(10^6+) running and don't want these to be erroring out necessarily
> 2. stetson_j_single has required parameters of fluxes and errors, meaning that a user would need to filter down to those band specific values manually for ingest into the function, would a single argument for band be better? This aligns it with stetson_J, but also would potentially make it redundant as it looks like stetson_J can handle a single input band string

1. I have placed a restriction that the flux array passed to _stetson_J_single has to have length larger than 1 or it returns np.nan.
2. I have renamed stetson_j_single to _stetson_j_single - I think this is really only a private function that user should never see, and all of the handling is via stetson_j functioning almost like a wrapper for a workhorse _stetson_j_single function

dougbrn (approved):
looks good

PR #8, opened 2022-11-16
Add basic unit testing

nevencaplar:
Add basic unit testing

This adds a simple test for StetsonJ function, using fixed data specified in the dictionary. The test is done on multiple python versions using github actions, ubunutu machine. The main drawback at this point is that it uses test.py in the code's main folder and manually specifies dependencies in the github action workflow.


.github/workflows/main.yml
```diff
@@ -0,0 +1,47 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    # Test the code on multiple python versions
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
```
drewoldag, line 19: Not a huge deal, but you probably don't need to run the build for all of these versions of Python. 3.8-3.10 are pretty standard. 3.11 is probably good, but basically brand new.

.github/workflows/main.yml
```diff
@@ -0,0 +1,47 @@
...
+          python3 -m pip install --upgrade pip
+          pip install numpy
+          pip install pandas
+
+      # Runs a single command using the runners shell
+      - name: Run a one-line script
+        run: echo Hello, world!
+
+      # Runs a set of commands using the runners shell
+      - name: Run a multi-line script
```
drewoldag, line 46: Probably want to update `name` here before it's integrated into CI pipeline. Just so that you have consistency in the future. Same story with lines 42 and 43.

drewoldag (approved):

.github/workflows/main.yml
nevencaplar: you mean something more informative and specific to this particular test, correct?

dougbrn:
lgtm, the only suggestion I might add is that perhaps our tests should live in their own folder?

dougbrn (approved):

.github/workflows/main.yml
drewoldag: Yeah that's right. Something like "Run unit tests"

nevencaplar:
> lgtm, the only suggestion I might add is that perhaps our tests should live in their own folder?

Yes, I wanted to do that - but then I got in trouble with importing a package in github actions. I will merge this as it is, and make a small issue for moving test.py into a dedicated folder

Issue #11, opened 2022-11-28
Pull analysis functions out of timeseries/ensemble


dougbrn:
For running timeseries through analysis functions, we'll want to support users going down two paths:

1. to_timeseries, working with a single or handful of objects closely
2. batch jobs, running analysis on many or all of the objects available in the ensemble

To support both of these, analysis functions should not live in timeseries or ensemble, but separately in their own module which ensemble and timeseries have wrapper functions for.

dougbrn:
Resolved in #12

Issue #10, opened 2022-11-28
Improve Indexing


dougbrn:
For timeseries and ensemble objects, an integer index is part of the multi-indexing scheme. This index uniquely indexes every row (counting from 0 onwards), but a likely improvement would be for it to uniquely index every (id,band) combination. Like the following:

id | band | index
01   |  u.  |     0
01   |  u.   |    1
01   |  u.   |    2
02  |   g.  |    0
01  |   g.  |     0
02  |   u.  |    0
02  |   u.   |    1

This helps to contextualize the index (it can be thought of as the nth observation of a given object in a given band)

dougbrn:
Resolved in #12

PR #12, opened 2022-11-30
Smart Indexing & Analysis Submodule

dougbrn:
This PR addresses two issues: #10 and #11 

Adds smart indexing; the lowest level index is an integer counting the nth row for each unique (id, band) pair. Previously the lowest level index just counted per row, without association to object id or band.

Adds an analysis submodule, and moves the stetsonJ function out of timeseries.py and into the submodule. This will enable groupby->apply style calls to these functions from the ensemble, which is a blocker for testing dask and vaex at scale.

A simple wrapper function for pandas.DataFrame.dropna was also added, but this will almost certainly be clobbered by whichever refactor route we go so I wouldn't spend too much time on it.

### DIFF
--- a/lsstseries/analysis/stetsonj.py
+++ b/lsstseries/analysis/stetsonj.py
@@ -1,0 +1,139 @@
+import numpy as np
+
+def calc_stetson_J(flux, err, band, band_to_calc=None):
+    """Compute the stetsonJ statistic on data from one or several bands
+        Parameters
+        ----------
+        flux : `numpy.ndarray` (N,)
+            Array of flux/magnitude measurements
+        err : `numpy.ndarray` (N,)
+            Array of associated flux/magnitude errors
+        band : `numpy.ndarray` (N,)
+            Array of associated band labels
+        band_to_calc : `str` or `list` of `str`
+            Bands to calculate stetson J on. Single band descriptor, or list 
+            of such descriptors.
+        Returns
+        -------
+        stetsonJ : `dict`
+            StetsonJ statistic for each of input bands.
+        Notes
+        ----------
+        In case that no value for `band_to_calc` is passed, the function is 
+        executed on all available bands in `band`.
+        """
+
+    unq_band = np.unique(band)
+
+    if band_to_calc is None:
+        band_to_calc = unq_band
+    if isinstance(band_to_calc,str):
+        band_to_calc = [band_to_calc]
+
+    assert hasattr(band_to_calc, '__iter__') is True
+
+    StetsonJ = {}
+    # TODO: ability to remove nan values
+    for b in band_to_calc:
+        if b in unq_band:
+            mask = band == b
+            fluxes = flux[mask]
+            errors = err[mask]
+            StetsonJ[b] = _stetson_J_single(fluxes, errors)
+        else:
+            StetsonJ[b] = np.nan
+
+    return StetsonJ
+
+def _stetson_J_single(fluxes, errors):
+    """Compute the single band stetsonJ statistic.
+    Parameters
+    ----------
+    fluxes : `numpy.ndarray` (N,)
+        Lightcurve flux values.
+    errors : `numpy.ndarray` (N,)
+        Errors on the lightcurve fluxes.
+    Returns
+    -------
+    stetsonJ : `float`
+        StetsonJ statistic
+    References
+    ----------
+    .. [1] Stetson, P. B., "On the Automatic Determination of Light-Curve
+    Parameters for Cepheid Variables", PASP, 108, 851S, 1996
+    Notes
+    ----------
+    Taken from
+    https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/diaCalculationPlugins.py
+    Using the function on random gaussian distribution gives result of -0.2
+    instead of expected result of 0?
+    """
+    mean = None
+    n_points = len(fluxes)
+    if n_points <= 1:
+        return np.nan
+    flux_mean = _stetson_mean(fluxes, errors, mean)
+    delta_val = (
+        np.sqrt(n_points / (n_points - 1)) * (fluxes - flux_mean) / errors)
+    p_k = delta_val ** 2 - 1
+    return np.mean(np.sign(p_k) * np.sqrt(np.fabs(p_k)))
+
+def _stetson_mean(values,
+                  errors,
+                  mean=None,
+                  alpha=2.,
+                  beta=2.,
+                  n_iter=20,
+                  tol=1e-6):
+    """Compute the stetson mean of the fluxes which down-weights outliers.
+    Weighted biased on an error weighted difference scaled by a constant
+    (1/``a``) and raised to the power beta. Higher betas more harshly
+    penalize outliers and ``a`` sets the number of sigma where a weighted
+    difference of 1 occurs.
+    Parameters
+    ----------
+    values : `numpy.dnarray`, (N,)
+        Input values to compute the mean of.
+    errors : `numpy.ndarray`, (N,)
+        Errors on the input values.
+    mean : `float`
+        Starting mean value or None.
+    alpha : `float`
+        Scalar down-weighting of the fractional difference. lower->more
+        clipping. (Default value is 2.)
+    beta : `float`
+        Power law slope of the used to down-weight outliers. higher->more
+        clipping. (Default value is 2.)
+    n_iter : `int`
+        Number of iterations of clipping.
+    tol : `float`
+        Fractional and absolute tolerance goal on the change in the mean
+        before exiting early. (Default value is 1e-6)
+    Returns
+    -------
+    mean : `float`
+        Weighted stetson mean result.
+    References
+    ----------
+    .. [1] Stetson, P. B., "On the Automatic Determination of Light-Curve
+    Parameters for Cepheid Variables", PASP, 108, 851S, 1996
+    Notes
+    ----------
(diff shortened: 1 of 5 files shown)


lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
...
+    if mean is None:
+        mean = np.average(values, weights=inv_var)
+    for iter_idx in range(n_iter):
+        chi = np.fabs(n_factor * (values - mean) / errors)
+        tmp_mean = np.average(
+            values,
+            weights=inv_var / (1 + (chi / alpha) ** beta))
+        diff = np.fabs(tmp_mean - mean)
+        mean = tmp_mean
+        if diff / mean < tol and diff < tol:
```
jeremykubica, line 137: There is a potential divide by zero here if mean is 0.0.

lsstseries/ensemble.py
```diff
@@ -24,7 +25,7 @@ def tap_token(self, token):
         """
         self.token = token
 
-    def query_tap(self, query, maxrec=None):
+    def query_tap(self, query, maxrec=None,debug=True):
```
jeremykubica, line 28: Style: Add a space after the comma.

lsstseries/ensemble.py
```diff
@@ -41,9 +42,14 @@ def query_tap(self, query, maxrec=None):
         """
         cred = vo.auth.CredentialStore()
         cred.set_password("x-oauth-basic", self.token)
-
+
```
jeremykubica, line 45: Style: extra space added.

lsstseries/ensemble.py
```diff
@@ -199,6 +205,17 @@ def _build_index(self, obj_id, band):
         # Create a multiindex
 
         idx = range(len(list(zip(obj_id, band))))
```
jeremykubica, line 207: You overwrite idx on line 211 below. Can you remove this line?

lsstseries/ensemble.py
```diff
@@ -199,6 +205,17 @@ def _build_index(self, obj_id, band):
...
+
+        #idx to count up for each unique pair of obj_id,band 
+        count_dict = {}
```
delucchi-cmu, line 210: I got stuck here, so I'm going to try to unpack my understanding and see if it matches the intent here.

You have might have (simplified) input like:
```
obj_id = [1,2,2,2,2,2,1]
band = [u,u,g,u,g,g,g]
```
Where you need those objects and bands to be the same length. you zip that up and get
`[1u,2u,2g,2u,2g,2g,1g]`

you want to end up with some nice indexed table that looks like
`[1u1,2u1,2g1,2u2,2g2,2g3,1g1]`

Do you need to maintain the order of the inputs? Because you could otherwise do something cute with

`obj_bands, counts = np.unique(zip(obj_id, band), axis=0, return_counts=true)`

That I think might read a little easier?

Either way, I think it would be helpful to have some comments here describing what your input expectations are, what mutations you're performing, and what the output should look like. Unit tests would also help document the expectations.

lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
...
+    .. [1] Stetson, P. B., "On the Automatic Determination of Light-Curve
+    Parameters for Cepheid Variables", PASP, 108, 851S, 1996
+    Notes
+    ----------
+    Taken from
+    https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/diaCalculationPlugins.py
+    """
+    n_points = len(values)
+    n_factor = np.sqrt(n_points / (n_points - 1))
+    inv_var = 1 / errors ** 2
```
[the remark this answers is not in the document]
nevencaplar: I would prefer not to touch and change the code that we ported fom the lsst pipeline itself. This is for two reasons:
1. The names of variables tend to be virtuall fixed and well understood in the lsst code
2. Introducing potential sources of confusion when we compare the results from running this code with the results derived from running origianl lsst code.

lsstseries/ensemble.py
nevencaplar: I concour. Linter run is needed.

lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
...
+        Parameters
+        ----------
+        flux : `numpy.ndarray` (N,)
+            Array of flux/magnitude measurements
+        err : `numpy.ndarray` (N,)
+            Array of associated flux/magnitude errors
+        band : `numpy.ndarray` (N,)
+            Array of associated band labels
+        band_to_calc : `str` or `list` of `str`
+            Bands to calculate stetson J on. Single band descriptor, or list
```
nevencaplar, line 14: Motivated by comment by Melissa - lets change this comment here to ``StetsonJ''.

lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+
+def calc_stetson_J(flux, err, band, band_to_calc=None):
```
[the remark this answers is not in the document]
nevencaplar: I think there is a mix of errors and valid different names here. I suggest

stetson_J - function name
StetsonJ - name for the statistics, in comments and such
stetsonJ - name for the dictionary

lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
...
+    unq_band = np.unique(band)
+
+    if band_to_calc is None:
+        band_to_calc = unq_band
+    if isinstance(band_to_calc,str):
+        band_to_calc = [band_to_calc]
+
+    assert hasattr(band_to_calc, '__iter__') is True
+
+    StetsonJ = {}
```
nevencaplar, line 35: As per previous discussion, I suggest changing dictionary name to stetsonJ

lsstseries/timeseries.py
```diff
@@ -93,7 +103,19 @@ def band(self):
     def _build_index(self, band):
         """Build pandas multiindex from band array"""
         # Create a multiindex
-        tuples = zip(band, range(len(band)))
+        idx = range(len(band))
+
+        #idx to count up for each unique pair of obj_id,band 
+        count_dict = {}
+        idx = []
```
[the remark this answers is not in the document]
dougbrn: Thanks for catching this, addressed in latest commit.

lsstseries/analysis/stetsonj.py
dougbrn: Done

lsstseries/analysis/stetsonj.py
dougbrn: Done

lsstseries/analysis/stetsonj.py
```diff
@@ -0,0 +1,139 @@
...
+    Taken from
+    https://github.com/lsst/meas_base/blob/main/python/lsst/meas/base/diaCalculationPlugins.py
+    Using the function on random gaussian distribution gives result of -0.2
+    instead of expected result of 0?
+    """
+    mean = None
+    n_points = len(fluxes)
+    if n_points <= 1:
+        return np.nan
+    flux_mean = _stetson_mean(fluxes, errors, mean)
```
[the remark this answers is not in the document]
dougbrn: Good catch, done

lsstseries/ensemble.py
dougbrn: Ran black on the three main files (ensemble, timeseries, stetsonj), first time using it but will be using it going forward so hopefully that will lower the frequency of my formatting issues 😅

lsstseries/ensemble.py
dougbrn: I think you have the right idea, though I'm running into an error getting your toy example to work. 
-> `AxisError: axis 0 is out of bounds for array of dimension 0`

Preserving the order of inputs is not a requirement at this point, but an improvement to this would be to order by timestamp most likely.

I'll link the issue here which has an example of what smart indexing wants to achieve input/output wise #10

I'll also mention here, that the motivation for this smart indexing is subtly influenced by a potential refactor to Xarray, and as such the same caveat for dropna applies here, this could be dropped if we go in the direction of Vaex or Dask.Dataframe as they don't support multi-indices.

nevencaplar (approved):
Looks good to me now. I also verified that the ensemble construction works as expected with an example.

lsstseries/ensemble.py
delucchi-cmu: I think it would be helpful to include the expected input/output alongside the method, either in docstring comments or unit tests that exercise the various options (preferably the unit tests).

dougbrn: A unit test is a good idea, I'll write one up later today

dougbrn: Unit test is in!

lsstseries/analysis/stetsonj.py
nevencaplar: I am torn about this one. On the one hand I would not want to touch ported lsst code. This is only a problem if values are 0, in which case mean is 0 as well. That should never be the case with the lsst data. 
Having said that, I am also uncomfortable about having this obvious point of possible failure here. One way to solve this is to add condition to if statement to break also if mean == 0?

dougbrn: +1 to Neven's suggestion, since the potential divide by zero is in a condition that just determines whether or not to break, adding another break condition seems appropriate.

Issue #23, opened 2023-01-19
Further enhancement of structure function and capability to calculate the result for ensemble


nevencaplar:
- Calculate structure function for an ensemble of lightcurves
	- sample the points so that each object has the same weight
	- sample all points equally - this has the problem that lightcurves that have more datapoints dominate the ensemble

- Different implementations of the structure function (by this, I mean different equations - reference for these different equations is here: TBD )

-  Combine points that have been taken during the same night - take mean, and increase errors

-  Simple filtering on the light curves
	- sigma clipping
	-``AGN specific’’ clipping - problem with sigma clipping is that it is not good for lightcurves that change brightness - I need to further specify this

dougbrn:
Additionally, from #22:

Add error estimate for structure function estimation via bootstrapping. Resample all of the points in a time bin and recompute the structure-function - do it e.g, 100 times and note 1-sigma limits.

nevencaplar:
This issue has been elaborated in better pointed https://github.com/lincc-frameworks/lsstseries/issues/59, https://github.com/lincc-frameworks/lsstseries/issues/60, https://github.com/lincc-frameworks/lsstseries/issues/61, https://github.com/lincc-frameworks/lsstseries/issues/62

Issue #22, opened 2023-01-19
Implement improvements to structure-function


nevencaplar:
-  Add error estimate for structure function estimation via bootstrapping. Resample all of the points in a time bin and recompute the structure-function - do it e.g, 100 times and note 1-sigma limits.

- Allow for the creation of automatic time bins in which structure function is calculated (similar as it is possible to do with Lomb-Scargle periodogram calculation). Add following options:
    - Time bins so that each time bin has the same number of points
    - Time bins so that each time bin has the same length of time
    - Time bins so that each time bin has same log length of time 

- Make sure that nan-s and other artifacts do not break the evaluation of the structure-function

- Implement non-dictionary output from the calculation; the output should be in a format that is easily accessible with standard python data manipulation methods.

dougbrn:
Marking this as resolved due to #25 , the request for bootstrapping will be added to the other issue

PR #25, opened 2023-02-02
First Set of Structure Function Enhancements

dougbrn:
This PR is the first set of enhancements to the structurefunction2 implementation in lsstseries. This addresses some of the requirements listed in #22 and #23, but I wanted to check in at this stage to make sure everything is on the right track.

What's in this PR:
* Ability to operate on many lightcurves, either independently in batch mode, or with combination to have multiple lightcurves contribute to the same result
* Some nan filtering, to make sf2 runnable in ensembles that have not had nans handled
* Improved output, a pandas dataframe is now returned. ensemble.batch has been improved to allow the dask meta keyword (which defines the output structure) to be passed, but also provides a baked-in solution for calc_sf2.
* Automated binning, implemented 3 options. "size" solves for bins of roughly equal size (number of dts in said bin), "length" provides bins of equal length, and "loglength" provides bin of equal length in log-space.

What remains to be done once this PR is merged in:
* Error estimates via bootstrapping
* sampling methods for multi-light curve case (when combine=True)
* Data point combination (in same night, etc.)
* filtering steps for the light curves

### DIFF
--- a/lsstseries/analysis/structurefunction2.py
+++ b/lsstseries/analysis/structurefunction2.py
@@ -4,7 +4,7 @@ import pandas as pd
 
 
 def calc_sf2(lc_id, time, flux, err, band,
-             band_to_calc=None, combine=False, sthresh=100):
+             band_to_calc=None, combine=False, method='size', sthresh=100):
     """Compute structure function squared on one or many bands
 
     Parameters
@@ -22,6 +22,15 @@ def calc_sf2(lc_id, time, flux, err, band,
     band_to_calc : `str` or `list` of `str`
         Bands to calculate structure function on. Single band descriptor,
         or list of such descriptors.
+    combine : 'bool'
+        Boolean to determine whether structure function is computed for each
+        light curve independently (combine=False), or computed for all light
+        curves together (combine=True).
+    method : 'str'
+        The binning method to apply, choices of 'size'; which seeks an even
+        distribution of samples per bin using quantiles, 'length'; which
+        creates bins of equal length in time and 'loglength'; which creates
+        bins of equal length in log time.
     sthresh : 'int'
         Target number of samples per bin.
 
@@ -66,7 +75,8 @@ def calc_sf2(lc_id, time, flux, err, band,
             fluxes_2d = [fluxes[mask] for mask in id_masks]
             errors_2d = [errors[mask] for mask in id_masks]
 
-            res = _sf2_single(times_2d, fluxes_2d, errors_2d, sthresh, combine=combine)
+            res = _sf2_single(times_2d, fluxes_2d, errors_2d, combine=combine,
+                              method=method, sthresh=sthresh)
 
             res_ids = [[str(unq_ids[i])]*len(arr) for i, arr in enumerate(res[0])]
             res_bands = [[b]*len(arr) for arr in res[0]]
@@ -82,7 +92,8 @@ def calc_sf2(lc_id, time, flux, err, band,
     return sf2_df
 
 
-def _sf2_single(times, fluxes, errors, sthresh, combine=False):
+def _sf2_single(times, fluxes, errors,
+                combine=False, method='size', sthresh=100):
     """Calculate structure function squared
 
     Calculate structure function squared from the available data. This is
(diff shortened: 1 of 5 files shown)


dougbrn:
@nevencaplar Don't worry about this until after your vacation

lsstseries/analysis/structurefunction2.py
```diff
@@ -41,20 +55,48 @@ def calc_sf2(time, flux, err, bins, band, band_to_calc=None):
 
     assert hasattr(band_to_calc, "__iter__") is True
 
-    sf2 = {}
+    ids = []
+    dts = []
+    bands = []
+    sf2s = []
     for b in band_to_calc:
```
delucchi-cmu, line 62: [Optional] I know that this is existing code, but it took me a moment to figure out what was going on here. These two lines are effectively a set intersection. I think the following logic is the same, and is a little more readable:

```
band_intersection = set(np.unique(band))
if band_to_calc:
  band_intersection = band_intersection.intersection(band_to_calc) // this is ok with set or single string

for b in band_intersection:
  // do stuff
```

tests/test_ensemble.py
```diff
@@ -79,3 +80,22 @@ def test_build_index():
     result = list(ens._build_index(obj_ids, bands).get_level_values(2))
     target = [0, 1, 2, 0, 0, 0, 1]
     assert result == target
+
+
+@pytest.mark.parametrize("method", ["size", "length", "loglength"])
```
delucchi-cmu, line 85: NICE!

lsstseries/analysis/structurefunction2.py
```diff
@@ -1,38 +1,52 @@
 from scipy.stats import binned_statistic
 import numpy as np
+import pandas as pd
 
 
-def calc_sf2(time, flux, err, bins, band, band_to_calc=None):
+def calc_sf2(lc_id, time, flux, err, band,
+             band_to_calc=None, combine=False, method='size', sthresh=100):
```
nevencaplar, line 7: It should still be possible to manually specify bins. I like the bins methods that you implemented, but they should not remove the possibility to manually add bins.

lsstseries/timeseries.py
```diff
@@ -162,7 +166,9 @@ def sf2(self, bins=None, band=None):
 
         Notes
         ----------
-        In case that no value for band is passed, the function is executed
+        In case that no value for band_to_calc is passed, the function is executed
         on all available bands.
         """
-        return calc_sf2(self.time, self.flux, self.flux_err, bins, self.band, band_to_calc=band)
+        lc_id = [self.meta['id']]*len(self.time)
```
nevencaplar, line 172: I tried testing code with a timeseries that I created from a dictionary (initial way we did it) and that fails because it has no id value. Can we make that it is possible to calculate sf2 for a single timeseries without id value? 

I see that in the test you manually insert id value - I am fine with that as long as it is hidden from the user.

lsstseries/ensemble.py
```diff
@@ -434,3 +452,47 @@ def _build_index(self, obj_id, band):
         tuples = zip(obj_id, band, idx)
         index = pd.MultiIndex.from_tuples(tuples, names=["object_id", "band", "index"])
         return index
+
+    def sf2(self, band_to_calc=None, combine=False,
+            method="size", sthresh=100):
```
nevencaplar, line 457: What happens if user passes a lightcurve that has less than 100 pairs?

lsstseries/analysis/structurefunction2.py
dougbrn: Seems reasonable, will do

lsstseries/timeseries.py
dougbrn: We should expect to have an ID populated through most of our timeseries ingest methods eventually, but agree that there's no downside to having it work when it's not specified here.

lsstseries/ensemble.py
dougbrn: Whenever sthresh exceeds the total number of pairs, it simply lumps all pairs into a single bin. This should be true for all three methods.

nevencaplar: I suggest adding a warning that the number of pairs is lower than the treshold - either lower the threshold or increase the number of data points.

dougbrn: My concern with that is that warnings won't scale, e.g. if we batch run 1 billion objects and even 1000 have this issue the output warnings are just spam and not useful

nevencaplar (approved):
Looks good to me now.

lsstseries/analysis/structurefunction2.py
dougbrn: Nice, this streamlines it quite a bit. Implemented

dougbrn: Err, reverted back. This was causing the CI to fail. Not entirely sure why, perhaps the fact that sets in Python are not ordered? Don't want this to hold things up.

PR #32, opened 2023-02-13
README tweaks for installation.

drewoldag:
Added python version to `conda create` so pip will be installed, and added `pip install pytest` for testing.


dougbrn (approved):
looks good, thanks Drew!

PR #33, opened 2023-02-13
Incorporating python-project-template into existing work. 

drewoldag:
The steps that I took to do this work are outlined in this doc: 
https://docs.google.com/document/d/1dpVupgOB0saF1e52UvJM3R72dHKanF2TXVuiceF9Yho/edit


drewoldag:
Looks like the failed CI steps are due to pylint. This doesn't come as a big surprise, because in order to commit/push without fixing all the pylint "errors", I specifically ignored those by using: `SKIP=pylint git commit -m 'blah'`.

docs/conf.py
```diff
@@ -0,0 +1,59 @@
...
+# Define path to the code to be documented **relative to where conf.py (this file) is kept**
+sys.path.insert(0, os.path.abspath('../src/'))
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'lsstseries'
+copyright = '2023, Doug Branton'
+author = 'Doug Branton'
+release = 'v0.1'
```
delucchi-cmu, line 21: [Aside] Should this use the project's version, or is the doc versioning distinct from library versioning?

docs/index.rst
```diff
@@ -0,0 +1,13 @@
+.. lsstseries documentation master file, created by
```
delucchi-cmu, line 1:
```suggestion
.. lsstseries documentation main file.
```

delucchi-cmu (approved):
Mostly pointed out things that are not specific to this repo, but I noticed while reviewing. Feel free to resolve them, or start an offline conversation.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,83 @@
...
+        language: system
+        entry: jupyter nbconvert --ClearOutputPreprocessor.enabled=True --inplace
+
+    # Run unit tests, verify that they pass
+  - repo: local
+    hooks:
+      - id: pytest-check
+        name: pytest-check
+        description: Run unit tests with pytest.
+        entry: pytest --cov=./src --cov-report=html
```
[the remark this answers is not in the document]
drewoldag: Yes! Good eye, spotting the  difference here. Since the `./src` directory contains the code that will be committed, we want to run the coverage report against that.

When the workflow in github runs, the code will be installed into site-packages first. So in that case, we want to use `--cov=<package_name>`.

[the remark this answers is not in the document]
drewoldag: I've added a comment to the python-package-template in the this file to explain the difference.

docs/conf.py
drewoldag: No, I don't think that the documentation versioning should be distinct from the code versioning. Good call here.

docs/conf.py
```diff
@@ -0,0 +1,59 @@
...
+
+# Define path to the code to be documented **relative to where conf.py (this file) is kept**
+sys.path.insert(0, os.path.abspath('../src/'))
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'lsstseries'
+copyright = '2023, Doug Branton'
+author = 'Doug Branton'
```
dougbrn, line 20: Are multiple authors supported here? Would be good to have Neven here if possible.

.gitignore
```diff
@@ -1,14 +1,133 @@
...
+docs/_build/
+
+# PyBuilder
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
 
-# dask
-dask-worker-space/
```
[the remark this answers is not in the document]
drewoldag: Add this to the python-project-template in this PR: https://github.com/lincc-frameworks/python-project-template/pull/56

docs/conf.py
drewoldag: Yeah, the template only adds one, but you should be able to add more. Pretty sure that the author and copyright info are just strings. So you should be able to do:
`author = 'Doug Branton, Neven Caplar'`

.pre-commit-config.yaml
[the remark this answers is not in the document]
drewoldag: I've added this to the python-project-template in this PR: https://github.com/lincc-frameworks/python-project-template/pull/56

docs/conf.py
drewoldag: I've added support for using `setuptools_scm` here in python-project-template with this PR: https://github.com/lincc-frameworks/python-project-template/pull/56

dougbrn (approved):

PR #36, opened 2023-02-16
Inject dask client into ensemble

delucchi-cmu:
- Allows for a single dask client to be created for all unit tests. This speeds up execution and reduces warnings.
- Locally, unit test execution goes from 35-45 seconds to 8-9 seconds
- In github CI, execution goes from 67-77 seconds to 16-19 seconds
- Addresses nearby black formatting and pylint warnings

### DIFF
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -12,6 +12,8 @@ from .timeseries import timeseries
 class ensemble:
     """ensemble object is a collection of light curve ids"""
 
+    client = None
+
     def __init__(self, token=None, client=None, **kwargs):
         self.result = None  # holds the latest query
         self.token = token
@@ -27,8 +29,16 @@ class ensemble:
         # Setup Dask Distributed Client
         if client:
             self.client = client
-        else:
+        elif not self.client:
             self.client = Client(**kwargs)  # arguments passed along to Client
+
+    def __enter__(self):
+        self.client = Client()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.client.close()
+        return self
 
     def client_info(self):
         """Calls the Dask Client, which returns cluster information
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -12,8 +12,6 @@ from .timeseries import timeseries
 class ensemble:
     """ensemble object is a collection of light curve ids"""
 
-    client = None
-
     def __init__(self, token=None, client=None, **kwargs):
         self.result = None  # holds the latest query
         self.token = token
@@ -26,18 +24,26 @@ class ensemble:
         self._err_col = "psFluxErr"
         self._band_col = "band"
 
+        self.client = None
+        self.cleanup_client = False
         # Setup Dask Distributed Client
         if client:
             self.client = client
-        elif not self.client:
+        else:
             self.client = Client(**kwargs)  # arguments passed along to Client
+            self.cleanup_client = True
 
     def __enter__(self):
-        self.client = Client()
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        self.client.close()
+        if self.cleanup_client:
+            self.client.close()
+        return self
+    
+    def __del__(self):
+        if self.cleanup_client:
+            self.client.close()
         return self
 
     def client_info(self):
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -12,20 +12,23 @@ from .timeseries import timeseries
 class ensemble:
     """ensemble object is a collection of light curve ids"""
 
-    def __init__(self, token=None, **kwargs):
+    def __init__(self, token=None, client=None, **kwargs):
         self.result = None  # holds the latest query
         self.token = token
         self.data = None
 
         # Assign Default Values for critical column quantities
-        self._id_col = 'object_id'
-        self._time_col = 'midPointTai'
-        self._flux_col = 'psFlux'
-        self._err_col = 'psFluxErr'
-        self._band_col = 'band'
+        self._id_col = "object_id"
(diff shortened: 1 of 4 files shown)


src/lsstseries/ensemble.py
```diff
@@ -12,20 +12,23 @@
...
+        self._flux_col = "psFlux"
+        self._err_col = "psFluxErr"
+        self._band_col = "band"
 
         # Setup Dask Distributed Client
-        self.client = Client(**kwargs)  # arguments passed along to Client
+        if client:
+            self.client = client
+        else:
+            self.client = Client(**kwargs)  # arguments passed along to Client
```
smcguire-cmu, line 31: I like the injecting client design, but I think if we're doing that we might not want to have the class making its own client. Or if we are keeping it, can we add `__enter__` and `__exit__` methods or a stop method that closes the client?

tests/lsstseries_tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+
+from lsstseries import ensemble
+
+
+@pytest.fixture(scope="package", name="dask_client")
+def dask_client():
+    """Create a single client for use by all unit test cases."""
+    client = Client()
+    yield client
+    client.close()
```
smcguire-cmu, line 13: This is great :)

smcguire-cmu (commented):
Looks really good mostly, just want to check about the client creation in the class

src/lsstseries/ensemble.py
delucchi-cmu: Fair. My motivation was to not make too many intrusive changes to this class, since I don't have full context on other ways this might get called or when a client might be necessary. Hopefully Doug can weigh in too!

dougbrn: The motivation for ensemble creating its own client is from a usability standpoint. It's reasonable to expect most of our users are not familiar with dask, so having the ensemble create a client if one is not provided seems most accessible. Open to hearing alternatives though, and in favor of the` __enter__` and `__exit__` methods in the case where this is kept

tests/lsstseries_tests/test_analysis.py
```diff
@@ -40,21 +42,21 @@ def test_sf2_timeseries(lc_id):
         "flux_err": test_yerr,
         "band": ["r"] * len(test_y),
     }
-    ts = timeseries()
-    ts.meta['id'] = lc_id
-    test_ts = ts.from_dict(data_dict=test_dict)
-    res = test_ts.sf2(sthresh=100)
+    timseries = timeseries()
```
dougbrn, line 45: small thing, but I think my preference is the "ts" shorthand over "timseries", more akin to pandas running with "df" as its default shorthand for a dataframe. It doesn't matter for these tests really, but in documentation we'll most likely be assigning this as "ts".

tests/lsstseries_tests/test_ensemble.py
```diff
@@ -5,98 +5,91 @@
...
 
 
-def test_to_timeseries(parquet_data):
+def test_to_timeseries(parquet_ensemble):
     """
     Test that ensemble.to_timeseries() runs and assigns the correct metadata
     """
-    ens = parquet_data
-    ts = ens.to_timeseries('88480000290704349')
+    timeseries = parquet_ensemble.to_timeseries(88480000290704349)
```
dougbrn, line 62: similarly here, regarding naming, if we ever need to actually bring in the timeseries module this will conflict with the namespace

tests/lsstseries_tests/test_analysis.py
delucchi-cmu: Reverted.

pylint was complaining about short variable names, but repo convention trumps pylint!

src/lsstseries/ensemble.py
```diff
@@ -12,6 +12,8 @@
 class ensemble:
```
smcguire-cmu, line 12: I know this isn't changed in this PR, but can we capitalize this class name?

src/lsstseries/ensemble.py
delucchi-cmu: `__enter__` and `__exit__` are only called when initialized in a `with` statement, so the most recent change means there are three ways a client could be created for an ensemble:

```
with Client(**kwargs) as client:
  ens = ensemble(client=client)
```
or 
```
with ensemble() as ens:
  ens.do_stuff()
```
or the most plain way:
```
ens = ensemble()
```

tests/lsstseries_tests/test_ensemble.py
```diff
@@ -5,7 +5,19 @@
 from lsstseries.analysis.stetsonj import calc_stetson_J
 from lsstseries.analysis.structurefunction2 import calc_sf2
 
+
 # pylint: disable=protected-access
+def test_with():
+    """Test that we open and close a client on enter and exit."""
+    with ensemble() as ens:
```
smcguire-cmu, line 12: I know this isn't from this pull request, but can we change the class name to be capitalized?

I'm happy to make a different issue and do it myself if we don't want to here.

src/lsstseries/ensemble.py
smcguire-cmu: I think this is good. Though it might also be good to add an `ensemble.close()` method to allow users to manage it themselves similar to `client.close()`

tests/lsstseries_tests/test_ensemble.py
dougbrn: Ah, yes, good eye @smcguire-cmu . @delucchi-cmu , up to you if you want to take this on here or just have this be it's own thing

delucchi-cmu: Happy to do in another PR. I was getting yelled at a lot by pylint about this one, but I was really trying not to step on too many toes!

dougbrn (approved):

PR #43, opened 2023-02-27
Map partitions

### LINKED ISSUE #19: Switch to map_partitions()
A map partitions call when batching analysis functions on ensemble can be implemented as such (in ensemble.batch):

`batch = self.data.map_partitions(lambda x: x.groupby(self._id_col).apply(lambda y: func(*[y[arg] for arg in args],
                                                         **kwargs)), meta=(self._id_col, type(self._id_col)))`

This will require that input data is partitioned such that one lightcurve is only ever in one partition, so the implementation of this is blocked by a partitioning tool that can satisfy that requirement. When implemented, this should improve performance noticeably, as it will eliminate data shuffling costs.

dougbrn:
By default, use map_partitions instead of an ensemble-wide groupby. Though the ensemble-wide groupby is still available to use if use_map is set to false. Map partitions creates more efficient task graphs and avoids data shuffling costs, but has the added requirement that each light curve is confined to a single partition.

Closes #19

### DIFF
--- a/tests/lsstseries_tests/test_ensemble.py
+++ b/tests/lsstseries_tests/test_ensemble.py
@@ -68,12 +68,13 @@ def test_prune(parquet_ensemble):
     assert parquet_ensemble.count(ascending=False).values[0] >= threshold
 
 
-def test_batch(parquet_ensemble):
+@pytest.mark.parametrize("use_map", [True, False])
+def test_batch(parquet_ensemble, use_map):
     """
     Test that ensemble.batch() returns the correct values of the first result
     """
     result = (
-        parquet_ensemble.prune(10).dropna(1).batch(calc_stetson_J, band_to_calc=None)
+        parquet_ensemble.prune(10).dropna(1).batch(calc_stetson_J, use_map=use_map, band_to_calc=None)
     )
 
     assert pytest.approx(result.values[0]["g"], 0.001) == -0.04174282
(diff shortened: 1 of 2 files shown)


tests/lsstseries_tests/test_ensemble.py
```diff
@@ -106,14 +107,14 @@ def test_build_index(dask_client):
...
+def test_sf2(parquet_ensemble, method, combine, sthresh, use_map=False):
     """
     Test calling sf2 from the ensemble
     """
 
-    res_sf2 = parquet_ensemble.sf2(combine=combine, method=method, sthresh=sthresh)
+    res_sf2 = parquet_ensemble.sf2(combine=combine, method=method, sthresh=sthresh, use_map=use_map)
     res_batch = parquet_ensemble.batch(
-        calc_sf2, combine=combine, method=method, sthresh=sthresh
+        calc_sf2, use_map=False, combine=combine, method=method, sthresh=sthresh
```
nevencaplar, line 117: Is there a reason why this is explicitly `False` and not using the argument supplied `use_map`

dougbrn: Only if you count it being a mistake as a reason 😅 , good catch. It's been fixed now

nevencaplar (approved):
Looks good to me.

PR #45, opened 2023-03-09
add nan filtering

dougbrn:
Add option to filter nans in calc_stetson_J. It's available in calc_sf2, and so the same functionality should be available in calc_stetson_J


nevencaplar (approved):
Looks good to me.

Issue #46, opened 2023-03-09
calc_sf2 pandas.qcut bottleneck


dougbrn:
Dask profiling on batch runs of calc_sf2 suggests that pandas.qcut is taking over half of the compute time. Should investigate this and potential replacements.

![Screen Shot 2023-03-09 at 12 56 19 PM](https://user-images.githubusercontent.com/8853505/224156404-d443eb9d-5089-440a-9a2a-c044611b629f.png)

jeremykubica:
Do you need to remove duplicates? If not, I think the following code work in qcuts place:

`num_quantiles = int(np.ceil(len(dts) / sthresh))`
`quantiles = np.linspace(0.0, 1.0, num_quantiles + 1)`
`bins = np.quantile(dts, quantiles)`
`return bins`

I timed it out and the numpy version is about 8x faster.

jeremykubica:
At least partially helped by #52

PR #47, opened 2023-03-09
convert series to numpy arrays in batch function

dougbrn:
Ensemble.batch passes the data columns along to analysis functions as pandas.series objects, as our analysis functions are designed to consider 1d array-like inputs, this can introduce a sizeable slowdown. By casting these series objects to numpy arrays, we see speedups of ~10x when testing on calc_stetson_J. Calc_sf2 does not benefit from the change, but that's potentially because of #46


nevencaplar (approved):
Nice! Looks good to me.

PR #49, opened 2023-03-14
copier run, using previous answers

dougbrn:
This run is using the existing copier answers


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,33 @@
...
+    strategy:
+      matrix:
+        python-version: ['3.8', '3.9', '3.10']
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python {% raw %}${{ matrix.python-version }}{% endraw %}
+      uses: actions/setup-python@v4
+      with:
+        python-version: {% raw %}${{ matrix.python-version }}{% endraw %}
```
drewoldag, line 23: I just created this PR in the python project template to fix this: https://github.com/lincc-frameworks/python-project-template/pull/115/files

But basically, if you remove the `{% raw %}` and `{% endraw %}` tags on line 20 and 23, then this should become valid YAML, and the test should pass.

.pre-commit-config.yaml
```diff
@@ -48,19 +56,20 @@ repos:
...
+    rev: 23.1.0
+    hooks:
+      - id: black
+        types: [python]
+        files: ^(src|tests)/
+        # It is recommended to specify the latest version of Python
+        # supported by your project here, or alternatively use
+        # pre-commit's default_language_version, see
+        # https://pre-commit.com/#top_level-default_language_version
+        language_version: python3.10
```
drewoldag, line 71: nit: not sure if you want to update this to `python3.11`.

docs/conf.py
```diff
@@ -8,52 +8,39 @@
...
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
-project = 'lsstseries'
-copyright = '2023, Doug Branton, Neven Caplar'
-author = 'Doug Branton, Neven Caplar'
-release = 'v0.1'
+project = "lsstseries"
+copyright = "2023, Doug Branton"
+author = "Doug Branton"
```
drewoldag, line 21: Looks like Neven's name was dropped on lines 20 and 21.

drewoldag (approved):
I left a couple small comments, but otherwise looks good to me.

docs/conf.py
dougbrn: Changed to use the new lincc mailing list

PR #50, opened 2023-03-20
Init docs

dougbrn:
This PR adds a set of tutorial notebooks that layout the current capabilities of lsstseries and show the basics of how users would use the package. In particular, the notebooks outlining the Ensemble object and scaling to large data are most important. The notebook showing the TimeSeries object is very initial and I wouldn't worry too much about reviewing it unless something really sticks out.


docs/notebooks/working_with_the_timeseries.ipynb
```diff
@@ -0,0 +1,112 @@
...
+    "As a result, we get a multi-indexed `Pandas` DataFrame with data from a single lightcurve. The multi-index contains a band index as well as a integer index. We can operate on this as we normally would a `Pandas` DataFrame."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import matplotlib.pyplot as plt\n",
```
drewoldag, line 68: I noticed that matplotlib isn't listed as a requirement in pyproject.toml. But after I installed it, this worked as expected.

drewoldag (approved):
All of this looks good to me! The only comment was about a missing dependency on matplotlib.

docs/notebooks/working_with_the_timeseries.ipynb
dougbrn: Yeah, it's not needed at all for the code itself, but it is for if you want to run the notebooks locally. Not sure what the right call there is as to whether include it or not.

drewoldag: One approach is to manually keep your docs/requirements.txt in sync with your pyproject.toml 'dev' dependencies section. You could also create a separate dependencies section in pyproject.toml, something like 'docs' that would just have dependencies needed to build notebooks that might not be necessary for other dev work. 

Or neither approach is fine too. As long as the docs/requirements.txt file is correct, then Sphinx/GitHub/readthedocs will be able to build and render the notebooks.

PR #52, opened 2023-03-23
Replace the Pandas cut and qcut with numpy functions

jeremykubica:
Replacing `qcut` with numpy `quantile` leads to a >8x speedup. Since we are only using the bins from the `cut` function, we can replace those with a `linspace`.

There is some extra logic to ensure this is a complete no-op (pandas extends the lower bound of the fit bin in the cut function by 0.1% of the range). I'm not sure if we need to keep that or not. But including it right now for consistency.


dougbrn (approved):
This looks great, I tested this change via a few batch jobs of ztf on epyc and found that this improved the overall speed of calc_sf2 by a factor of ~4x, which is wonderful.

PR #56, opened 2023-03-30
Small optimization

jeremykubica:
Small (and probably not necessary) optimization to remove conversion to string during index building. Added a few more cases to the test as well.


dougbrn (approved):
looks good, thanks!

PR #57, opened 2023-03-30
Remove string formats from TimeSeries _build_index

jeremykubica:
Another very small optimization. I also added some tests focused just on `TimeSeries`.


dougbrn (approved):
nice! Looks good to me.

Issue #75, opened 2023-04-11
Persist objects that have no sources


dougbrn:
Currently, if all sources tied to a given object are removed from the source table, the object will be removed from the object table as well. This is because the sync operation from source to object generates a new object table from the source table and merges the existing table in, keeping only the ids found in the new object table. The ideal solution would be to keep those objects in the object table with the nobs columns set to 0. We should add this to the ensemble.

PR #80, opened 2023-04-13
Give _sync_tables the ability to persist objects that have no sources 

jeremykubica:
This PR addresses issue #75. Creates an optional parameter that allows `_generate_object_table` to keep any object ids are no longer in the `_source` table but were previously in the `_object` table. It populates all counts for these objects with zero.

### DIFF
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -483,10 +483,10 @@ class Ensemble:
         # objects, then copy them into the res table with counts of zero.
         if self.keep_empty_objects and self._object is not None:
             # Check that there are existing object ids.
-            object_inds = self._object.index.values.compute()
+            object_inds = self._object.index.unique().values.compute()
             if len(object_inds) > 0:
                 # Determine which object IDs are missing from the source table.
-                source_inds = self._source.index.values.compute()
+                source_inds = self._source.index.unique().values.compute()
                 missing_inds = np.setdiff1d(object_inds, source_inds).tolist()
 
                 # Create a dataframe of the missing IDs with zeros for all bands and counts.
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -482,6 +482,8 @@ class Ensemble:
         # If the ensemble's keep_empty_objects attribute is True and there are previous
         # objects, then copy them into the res table with counts of zero.
         if self.keep_empty_objects and self._object is not None:
+            prev_partitions = self._object.npartitions
+
             # Check that there are existing object ids.
             object_inds = self._object.index.unique().values.compute()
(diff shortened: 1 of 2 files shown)


src/lsstseries/ensemble.py
```diff
@@ -475,27 +482,52 @@ def _generate_object_table(self):
...
-        # Add total nobs
-        res[self._nobs_col] = counts.groupby(self._id_col).agg("sum")
+        # Add total nobs by summing across each band.
+        res[self._nobs_col] = res.sum(axis=1)
 
         return res
 
-    def _sync_tables(self):
-        """Sync operation to align both tables"""
+    def _sync_tables(self, keep_zero_entries=False):
```
dougbrn, line 511: With this as a kwarg for a hidden function, should we add an ensemble flag to control this? Ensemble.keep_empty_objects or something? If we do, small additional note that the first generation of table from source in read_parquet will always want keep_zero_entries set to False, regardless of the flag, as we wouldn't have an existing ensemble._object to compare to.

jeremykubica: That's a really good call. I was wondering what the best way to handle turning it off and on (and clearly decided to hold off making that decision). I'm happy to change the code to enable by flag though.

jeremykubica: I've updated the code.

src/lsstseries/ensemble.py
```diff
@@ -475,21 +479,46 @@ def _generate_object_table(self):
...
+                # Create a dataframe of the missing IDs with zeros for all bands and counts.
+                rows = {self._id_col: missing_inds}
+                for i in res.columns.values:
+                    rows[i] = [0] * len(missing_inds)
+
+                zero_pdf = pd.DataFrame(rows, dtype=int).set_index(self._id_col)
+                zero_ddf = dd.from_pandas(zero_pdf, sort=True, npartitions=1)
+
+                # Concatonate the zero dataframe onto the results.
+                res = dd.concat([res, zero_ddf], interleave_partitions=True).astype(int)
```
dougbrn, line 501: My only other comment is how this creates a new partition for the zero source objects. I _think_ the preferred option for now would be to append these to the last partition? My thinking being this is the object table, which is generally much smaller than the source table. In many cases I think users will only need a single partition for their object table, so appending them would preserve that. For cases where we need many partitions for the object table I think we would eventually have them being appended by division, where each index is slotted into an existing partition using the divisions as guidelines. But we're not there yet on that end. Just my thinking on this though.

src/lsstseries/ensemble.py
dougbrn: Nice, looks good!

dougbrn (approved):
Looks great, thanks Jeremy!

PR #82, opened 2023-04-17
Rename the _from_ensemble function

jeremykubica:
Renames `TimeSeries._from_ensemble()` to `TimeSeries.from_dataframe()` since the function takes a `DataFrame` instead of an `Ensemble`.

Also adds parameters to the function's description and adds a test.


dougbrn (approved):
Nice, looks good to me!

PR #84, opened 2023-04-17
Allow users to not pass timestamps into SF2 function

drewoldag:
The actual logic here is relatively small (and so hopefully safe). Most of this PR is adding unit tests.

### DIFF
--- a/src/lsstseries/analysis/structurefunction2.py
+++ b/src/lsstseries/analysis/structurefunction2.py
@@ -13,7 +13,8 @@ def calc_sf2(
     lc_id : 'numpy.ndarray' (N,)
         Array of lightcurve ids per data point.
     time : `numpy.ndarray` (N,)
-        Array of times when measurements were taken.
+        Array of times when measurements were taken. If all array values are
+        `None`, then equidistant times between measurements are assumed.
     flux : `numpy.ndarray` (N,)
         Array of flux/magnitude measurements.
     err : `numpy.ndarray` (N,)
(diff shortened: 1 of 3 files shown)


src/lsstseries/analysis/structurefunction2.py
```diff
@@ -69,6 +70,13 @@ def calc_sf2(
 
             # Mask on band
             times = np.array(time)[band_mask]
+
+            # if times are all `None`, we assume equidistant times between
+            # measurements. To do so, we'll create an array of integers
+            # from 0 to N where N is the number of flux values for this band.
+            if np.all(np.equal(times, None)):
```
jeremykubica, line 77: Small comment and it is fine to leave it as-is: Alternatively (or in addition) you could check if `times` itself is None. This means that the user wouldn't need to allocate an array of Nones.

drewoldag: I was a little on the fence about supporting that use case, based only the existing docstring. But I do think it's a good idea. So I'll work in the changes necessary to support that.

tests/lsstseries_tests/test_analysis.py
```diff
@@ -83,3 +133,59 @@ def test_dt_bins():
...
+        combine=False,
+        method="size",
+        sthresh=100,
+    )
+
+    assert res["dt"][0] == pytest.approx(3.705, rel=0.001)
+    assert res["sf2"][0] == pytest.approx(0.005365, rel=0.001)
+
+
+def test_sf2_base_case_time_as_none_array():
```
drewoldag, line 166: A little extra code coverage, but unrelated to this work.

dougbrn (approved):
This looks good to me, especially with support for times=None now

PR #86, opened 2023-04-18
Update structure function docstrings to match DM recommendations

drewoldag:


src/lsstseries/analysis/structurefunction2.py
```diff
@@ -10,9 +10,9 @@ def calc_sf2(
 
     Parameters
     ----------
-    lc_id : 'numpy.ndarray' (N,)
+    lc_id : `numpy.ndarray` (N,)
         Array of lightcurve ids per data point.
-    time : `numpy.ndarray` (N,) or None
+    time : `numpy.ndarray` (N,) or `None`
```
nevencaplar, line 15: Is this how it is implemented. You need to specify None for times? What happens when you pass only a single array of values?

dougbrn, line 15: I'm also just realizing that at present this is hard to actually call Ensemble.batch() on. The reason being that batch passes along the time column directly to the function, there isn't a great way to signal to batch that you want to overwrite one of the input columns with your own input. A potential solution to this would be to also add a `use_times` boolean kwarg as well, so that the user could have times be populated but have a kwarg switch to force the assumption of equal spacing. It looks a little nonsensical from the perspective of solely calc_sf2, but without it users would have a hard time actually accessing this feature from the ensemble. Let me know what others think, if this is scope creep here I can also just make another issue

drewoldag: At present, yes, this is how it's implemented. A user would pass either `None` or an array of timestamps. Within the array of timestamps, some or all of the elements could be `None`. 
For instance, these are currently valid inputs:

- `None`
- `[None, None, None, None, ...]` 
- `[1, None, 4, 11, None, ...]`
- `[1, 2, 5, 11, ...]`


Perhaps I'm not understanding your question entirely. When you say "a single array of values", do you mean something like this:
` [1, 2, 5, 11, ...]` 
or something like this:
`[2, 2, 2, 2, ...]`

For the prior case, the function works as expected. For the later case, I believe the function would either fail or produce nonsensical output.

drewoldag: I see what you're talking about in `Ensemble.batch`. The solution you're proposing makes sense. I would like to think about it a little more, and perhaps chat it out some. 

I am curious if we plan on supporting both `Ensemble.batch` to call SF2 as well as `Ensemble.sf2`. Is there a subtle (or not so subtle :)) difference between the two? It seems that `batch` is made to be more general purpose, but I can't immediately tell.

dougbrn: If I'm remembering right, batch was made to be a pure "iterate over each lightcurve and apply a function" style function. The need for Ensemble.sf2 comes from the fact that calc_sf2 supports calculating one function from many lightcurves if the combine flag is set. So we came up with a specific loader for that use case, you can get a sense of that by looking at Ensemble.sf2 itself, where if combine is true it passes along a "mega-lightcurve", otherwise it just calls batch. https://github.com/lincc-frameworks/lsstseries/blob/main/src/lsstseries/ensemble.py#L849.

drewoldag: Looking over the code some more, I believe that I'm getting a better sense of what's going on. I'm assuming that `batch` here is leveraging Dask to schedule the operations. Where as `sf2` simply executes the `calc_sf2` function immediately. 

To that end, I assume that the concern expressed for `batch` also holds for `sf2` since there's no way to specify times there. Additionally, (please correct me if I've got this wrong), the same is true for `TimeSeries.sf2` The `sf2` method there doesn't allow for alternate time values. And it seems like you would have to recreate the TimeSeries in order to change the times (at least as it stands now)

dougbrn: Both Ensemble.batch and Ensemble.sf2 use Dask if combine is set to false (which is the case where you iterate over each lightcurve), combine set to true just directly calls calc_sf2 because it's just being executed once for the collection of all lightcurves. But generally I think you have the right idea. Everything implemented so far is right, we just may need the additional flexibility of a use_times kwarg so that the user can access this feature in batch calls. 
E.g.
```
# uses Ensemble._time_col values
Ensemble.batch(calc_sf2)  

# overrides Ensemble._time_col values to act as if None was supplied to calc_sf2
Ensemble.batch(calc_sf2, use_times=False)
```

drewoldag: Got it. That makes sense, and thanks for the confirmation.

drewoldag: So, not to be pushy, but is _this_ PR ok? I'm happy to open additional issues to track the work to make the modifications to `Ensemble` (and presumably `TimeSeries`).

nevencaplar: Regarding first part -> I would want that an user can pass only one list of values, and that then we assume it is a list of fluxes. I guess that is a problem because argument `flux` is not the first argument, you can not specify that `id` and `time` are optional, and `flux` is not? I.e., I would prefer if an user can avoid having to specify `time = None`, and could just invoke `calc_sf2([1,2,3, ...])`.

Regarding second part -> In both cases the function should work. I believe that in the case `[2, 2, 2, 2, ...]` the output should be a list of zeros (because the variance of the function is zero) but it should.

drewoldag: I believe I understand what you're saying now. Your preference is that the user would be able to invoke this function pass _just_ a list of fluxes and would assume equidistant times (and all errors = 0, etc...) If that assumption is correct, then yes, the function signature would either need to be refactored, or we would need to introduce an interface function on top that would allow that behavior. 

I'll put together a unit test to try out the [2,2,2,2...] scenario and see how the function responds. :)

PR #87, opened 2023-04-18
Allow user to pass `None`, scalar, or array as `err` parameter.

drewoldag:
This PR updates `calc_sf2` to support `None`, scalar, and arrays of values as input to the `err` parameter. When `None` is passed, we'll assume all measurement errors are 0. When a scalar is passed in, we'll assume the same error for all measurements. When an array is provided, we assume a distinct error for each measurement. 

There are a couple of ways to potentially optimize this that I didn't pursue, because I can imagine a refactor in the works soon. Namely, instead of generating arrays of either 0 or a repeated scalar value, we could pass a flag to the inner functions, signaling when to avoid extra work specifically for the case of `None` or scalar inputs. 

Also, as was mentioned by @dougbrn in a comment here: https://github.com/lincc-frameworks/lsstseries/pull/86#discussion_r1170499483 The new functionality introduced in this PR won't be accessible to `Ensemble`s via either `Ensemble.batch` or `Ensemble.sf2`, nor to `TimeSeries.sf2` until we modify they way those classes interact with this analysis function.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -81,12 +81,15 @@ repos:
         exclude_types: [file, symlink]
         args:
           [
-            "-T", # Show full trace back on exception
-            "-E", # Don't use saved env. always read all files.
-            "-b", # Flag to select which builder to use
-            "html", # Use the HTML builder
-            "-d", # Flag for cached environment and doctrees
-            "docs/build/doctrees", # directory
+            "-M", # Run sphinx in make mode, so we can use -D flag later
+                  # Note: -M requires next 3 args to be builder, source, output
+            "html", # Specify builder
             "./docs", # Source directory of documents
-            "docs/build/html", # Output directory for rendered documents.
+            "./_readthedocs", # Output directory for rendered documents
+            "-T", # Show full trace back on exception
+            "-E", # Don't use saved env; always read all files
+            "-d", # Flag for cached environment and doctrees
+            "./docs/_build/doctrees", # Directory
+            "-D", # Flag to override settings in conf.py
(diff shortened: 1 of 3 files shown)


tests/lsstseries_tests/test_analysis.py
```diff
@@ -135,6 +135,19 @@ def test_dt_bins():
     assert len(bins) == 11
 
 
+def test_dt_bins_raises_exception():
```
drewoldag, line 138: This test should raise the code coverage to 100% for this file. But it otherwise unrelated to the core changes here.

.pre-commit-config.yaml
```diff
@@ -81,12 +81,15 @@ repos:
...
+            "-E", # Don't use saved env; always read all files
             "-d", # Flag for cached environment and doctrees
-            "docs/build/doctrees", # directory
-            "./docs", # Source directory of documents
-            "docs/build/html", # Output directory for rendered documents.
-          ]
\ No newline at end of file
+            "./docs/_build/doctrees", # Directory
+            "-D", # Flag to override settings in conf.py
+            "exclude_patterns=notebooks/*", # Exclude our notebooks from pre-commit
```
drewoldag, line 94: The changes here, should speed up the pre-commit checks that run the sphinx build. The primary change here is that jupyter notebooks will no longer be built with each commit. If you all were specifically wanting to check that the jupyter notebooks build for each commit, we should change this back. Also, we now specify the `./docs/_build/` directory for doctrees and `./_readthedocs` for html output. Without that change, we potentially end up in  one of those highly nested directory structure situations where Sphinx eventually fails.

dougbrn (approved):
Looks good to me

Issue #71, opened 2023-04-05
Bin data taken during given time-frame, default==same night


nevencaplar:
Implement binning function that combines data taken during the same night, in the same band; assuming there was no intrinsic change of luminosity during this time window and that these observations can be combined in order to decrease the uncertainty. Add an option to have custom duration of time over which data is averaged, larger or smaller than default value of ``1 night''.

nevencaplar:
1. I would not include any detection mechanism in this function. It is to the discretion of user to decide if it makes sense this. In general, e.g., in AGN the change of flux on one-night time scales is much smaller than the sensitivity so it makes sense to combine. But if you were hunting for supernovae it would not make sense.

2. Yes, I agree. Assume that errors are perfectly Gaussian and with sigmas stored in error column.

jeremykubica:
Closed with #90

PR #90, opened 2023-04-24
Bin data taken during given time-frame

jeremykubica:
Addresses issue #71. Bins the source data during a given time window.

### DIFF
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -314,7 +314,7 @@ class Ensemble:
         tmp_time_col = "tmp_time_for_aggregation"
         if tmp_time_col in self._source.columns:
             raise KeyError(f"Column '{tmp_time_col}' already exists in source table.")
-        ens._source[tmp_time_col] = self._source[ens._time_col].apply(
+        self._source[tmp_time_col] = self._source[self._time_col].apply(
             lambda x: np.floor(x / time_window) * time_window, meta=pd.Series(dtype=float)
         )
 
@@ -335,11 +335,14 @@ class Ensemble:
             for key in additional_cols:
                 aggr_funs[key] = additional_cols[key]
 
-        # Group the columns by id, band, and time bucket.
-        ens._source = ens._source.groupby([ens._id_col, ens._band_col, tmp_time_col])
+        # Group the columns by id, band, and time bucket and aggregate.
+        self._source = self._source.groupby([self._id_col, self._band_col, tmp_time_col]).aggregate(aggr_funs)
 
         # Fix the indices and remove the temporary column.
-        ens_.source = ens._source.reset_index().set_index(ens._id_col).drop(tmp_time_col, axis=1)
+        self._source = self._source.reset_index().set_index(self._id_col).drop(tmp_time_col, axis=1)
+
+        # Mark the source table as dirty.
+        self._source_dirty = True
 
     def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, **kwargs):
         """Run a function from lsstseries.TimeSeries on the available ids
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -284,18 +284,8 @@ class Ensemble:
 
         return self
 
-    def bin_sources(self, time_window=1.0, additional_cols=None, use_map=True, **kwargs):
+    def bin_sources(self, time_window=0.5, additional_cols=None, use_map=True, **kwargs):
         """Bin sources on within a given time range to improve the estimates.
-
-        Notes
-        -----
-        * This should only be used for slowly varying sources where we can
-        treat the source as constant within `time_window`.
-
-        * As a default the function only aggregates and keeps the id, band,
-        time, flux, and flux error columns. Additional columns can be preserved
-        by providing the mapping of column name to aggregation function with the
-        `additional_cols` parameter.
 
         Parameters
         ----------
@@ -309,6 +299,16 @@ class Ensemble:
             used (True). Using map_partitions is generally more efficient, but
             requires the data from each lightcurve is housed in a single
             partition. If False, a groupby will be performed instead.
+
+        Notes
+        -----
+        * This should only be used for slowly varying sources where we can
+        treat the source as constant within `time_window`.
+
+        * As a default the function only aggregates and keeps the id, band,
+        time, flux, and flux error columns. Additional columns can be preserved
+        by providing the mapping of column name to aggregation function with the
+        `additional_cols` parameter.
         """
         # Bin the time and add it as a column.
         tmp_time_col = "tmp_time_for_aggregation"
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -284,6 +284,63 @@ class Ensemble:
 
         return self
 
+    def bin_sources(self, time_window=1.0, additional_cols=None, use_map=True, **kwargs):
+        """Bin sources on within a given time range to improve the estimates.
+
+        Notes
+        -----
+        * This should only be used for slowly varying sources where we can
+        treat the source as constant within `time_window`.
+
+        * As a default the function only aggregates and keeps the id, band,
+        time, flux, and flux error columns. Additional columns can be preserved
+        by providing the mapping of column name to aggregation function with the
+        `additional_cols` parameter.
+
+        Parameters
+        ----------
+        time_window : `float` (optional)
+            The time range (in days) over which to consider observations in the same bin.
+        additional_cols : `dict` (optional)
+            A dictionary mapping column name to aggregation method.
+            Example: {"my_value_1": "mean", "my_value_2": "max"}
+        use_map : `boolean` (optional)
+            Determines whether `dask.dataframe.DataFrame.map_partitions` is
+            used (True). Using map_partitions is generally more efficient, but
+            requires the data from each lightcurve is housed in a single
+            partition. If False, a groupby will be performed instead.
+        """
+        # Bin the time and add it as a column.
+        tmp_time_col = "tmp_time_for_aggregation"
+        if tmp_time_col in self._source.columns:
+            raise KeyError(f"Column '{tmp_time_col}' already exists in source table.")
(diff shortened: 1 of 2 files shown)


src/lsstseries/ensemble.py
```diff
@@ -284,6 +284,66 @@ def prune(self, threshold=50, col_name=None):
...
+            used (True). Using map_partitions is generally more efficient, but
+            requires the data from each lightcurve is housed in a single
+            partition. If False, a groupby will be performed instead.
+        """
+        # Bin the time and add it as a column.
+        tmp_time_col = "tmp_time_for_aggregation"
+        if tmp_time_col in self._source.columns:
+            raise KeyError(f"Column '{tmp_time_col}' already exists in source table.")
+        self._source[tmp_time_col] = self._source[self._time_col].apply(
+            lambda x: np.floor(x / time_window) * time_window, meta=pd.Series(dtype=float)
```
nevencaplar, line 318: I believe we want the purest mean of the times when the observations were taken. I am not sure that I understand the need for np.floor?

src/lsstseries/ensemble.py
```diff
@@ -284,6 +284,66 @@ def prune(self, threshold=50, col_name=None):
 
         return self
 
+    def bin_sources(self, time_window=1.0, additional_cols=None, use_map=True, **kwargs):
+        """Bin sources on within a given time range to improve the estimates.
+
+        Notes
```
nevencaplar, line 290: In general the LSST code that I have seen has Notes at the bottom, after Parameters. Having said that, I think it is only a custom and not obligatory so I am fine if you leave it like this as well.

src/lsstseries/ensemble.py
```diff
@@ -284,6 +284,66 @@ def prune(self, threshold=50, col_name=None):
 
         return self
 
+    def bin_sources(self, time_window=1.0, additional_cols=None, use_map=True, **kwargs):
```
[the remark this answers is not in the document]
jeremykubica: Sounds good. I changed this to 0.5 days.

src/lsstseries/ensemble.py
```diff
@@ -284,6 +284,66 @@ def prune(self, threshold=50, col_name=None):
...
+        # Bin the time and add it as a column.
+        tmp_time_col = "tmp_time_for_aggregation"
+        if tmp_time_col in self._source.columns:
+            raise KeyError(f"Column '{tmp_time_col}' already exists in source table.")
+        self._source[tmp_time_col] = self._source[self._time_col].apply(
+            lambda x: np.floor(x / time_window) * time_window, meta=pd.Series(dtype=float)
+        )
+
+        # Set up the aggregation functions for the time and flux columns.
+        aggr_funs = {self._time_col: "mean", self._flux_col: "mean"}
```
dougbrn, line 322: Maybe more of a @nevencaplar question, is mean all we would need to support? When I think of binning I tend to think of a sum.

It also looks like the code is fairly rigid for this, since the error calculation depends on the aggregation done on flux. So I'm not neccesarily advocating for it to be generalized here, but was just curious if mean is best serving.

src/lsstseries/ensemble.py
jeremykubica: The floor is only used for binning the time. The temporary time column will contain floor(time/window)*window.

If we use `times = [i/10.0 for i in range(20)]` and `time_window = 0.5`, we get:

```
times = [0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 1.7, 1.8, 1.9]
times / 0.5 = [0.0, 0.2, 0.4, 0.6, 0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0, 2.2, 2.4, 2.6, 2.8, 3.0, 3.2, 3.4, 3.6, 3.8]
floor(times / 0.5) = [0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0, 2.0, 3.0, 3.0, 3.0, 3.0, 3.0]
floor(times / 0.5) * 0.5 = [0.0, 0.0, 0.0, 0.0, 0.0, 0.5, 0.5, 0.5, 0.5, 0.5, 1.0, 1.0, 1.0, 1.0, 1.0, 1.5, 1.5, 1.5, 1.5, 1.5]
```
So floor(times / 0.5) * 0.5 becomes are bin labels.

The actual times are computed as the mean of all times in the bin.

src/lsstseries/ensemble.py
```diff
@@ -284,6 +284,66 @@ def prune(self, threshold=50, col_name=None):
...
+        if self._err_col in self._source.columns:
+            aggr_funs[self._err_col] = dd.Aggregation(
+                name="err_agg",
+                chunk=lambda x: (x.count(), x.apply(lambda s: np.sum(np.power(s, 2)))),
+                agg=lambda c, s: (c.sum(), s.sum()),
+                finalize=lambda c, s: np.sqrt(s) / c,
+            )
+
+        # Add any additional aggregation functions
+        if additional_cols is not None:
```
dougbrn, line 334: What happens if one of (self._flux_col, self._time_col, self._err_col) are provided in additional_cols? It looks like it might clobber the assigned aggregations, which is kind of what I was asking about above, but this may not be intended?

src/lsstseries/ensemble.py
jeremykubica: Good question. The way the code is structured, you can actually overwrite the aggregation functions for time, flux, and error by including them in the additional columns dictionary (since those overwrite the items in aggr_funs). I hadn't thought of this as a use case, so I didn't include it in the docstring.

src/lsstseries/ensemble.py
jeremykubica: This was not originally my intention, but I think we should leave it in given your question above. I changed the variable name to reflect this and expanded the docstring.

src/lsstseries/ensemble.py
jeremykubica: See my note above about this binning method though. I'm not sure this is what we want vs something more dynamic.

src/lsstseries/ensemble.py
jeremykubica: I'd prefer to stay consistent with LSST code. I moved the notes section.

src/lsstseries/ensemble.py
dougbrn: I see you answered this above, but this may not be intended, if technically the error calculation should be dependent on the flux calculation. I could see someone using this for good, for example if they want custom aggregations for their errors. But also bad, as someone may update the flux aggregation and not realize the errors won't adapt. Maybe we should just throw a warning if the core aggregations would be overwritten?

src/lsstseries/ensemble.py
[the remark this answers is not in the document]
nevencaplar: Indeed,  58971.99 and 58972.01  should be grouped together. Any observation that is closer than 0.5 days to another observations should be grouped together.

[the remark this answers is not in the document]
jeremykubica: Let's talk about this at the meeting this afternoon. The definition of time window gets a bit tricky. If we bin any pair of observations that are within time_window, the bin size could grow very large. It wouldn't be a problem if we were working with a single survey, but if we combine data from multiple telescopes in different time zones, we could (in theory) get observations every hour for weeks continuously. There are a few ways to handle the binning that we can discuss.

hombit:
You can ignore it, but let me share my previous attempts implementing time binning.

One is from `light-curve` package, it allows to set custom time offset. The rendered docs for Rust version is [here](https://docs.rs/light-curve-feature/0.5.5/light_curve_feature/features/struct.Bins.html), fully vectorized Python implementation with numpy and scipy is [here](https://github.com/light-curve/light-curve-python/blob/b5034c6fc5fc6f2bbe5f444ac9c1a8a9abf24137/light-curve/light_curve/light_curve_py/features/bins.py#L24).

One more less clear attempt is [here](https://github.com/snad-space/snad/blob/55b3ffc813825521a4af5012aa157b52ef624cd9/snad/load/curves.py#L170), it has no time offset, but optionally could produce non-even grid for time by averaging it in the way similar to fluxes.

src/lsstseries/ensemble.py
```diff
@@ -285,19 +285,74 @@ def prune(self, threshold=50, col_name=None):
...
         """Bin sources on within a given time range to improve the estimates.
 
         Parameters
         ----------
-        time_window : `float` (optional)
+        time_window : `float`, optional
             The time range (in days) over which to consider observations in the same bin.
-        custom_aggr : `dict` (optional)
+        offset : `float`, optional
+            The offset in days to use for binning. This should correspond to noon
```
nevencaplar, line 347: I am not sure that we want to call that a `noon`. We are searching for longest gap in the observations, which in general should be somewhat like solar noon. At minimal I would emphasize that we are trying to _estimate_ solar noon.

src/lsstseries/ensemble.py
```diff
@@ -285,19 +285,74 @@ def prune(self, threshold=50, col_name=None):
...
+            return -1
+        return (best_mid_pt % 24.0) / 24.0
+
+    def bin_sources(self, time_window=1.0, offset=None, custom_aggr=None, use_map=True, **kwargs):
         """Bin sources on within a given time range to improve the estimates.
 
         Parameters
         ----------
-        time_window : `float` (optional)
+        time_window : `float`, optional
```
nevencaplar, line 344: I would be beneficial what is the default value when the input is not specified =

src/lsstseries/ensemble.py
jeremykubica: Good call. I clarified the comment.

src/lsstseries/ensemble.py
jeremykubica: Done.

nevencaplar (approved):

dougbrn (approved):
This looks good to me, my one comment is that I think there's enough going on here to motivate an example or two being added to the documentation/tutorial notebooks. I don't think that needs to happen in this PR, but we should capture this in an issue.

PR #91, opened 2023-04-24
Add ability to load an ensemble from a dictionary

jeremykubica:
I expect this function will mostly be used for testing, but it might also be used for one-off notebook style analysis.

### DIFF
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -486,6 +486,65 @@ class Ensemble:
 
         return self
 
+    def from_source_dict(
+        self,
+        source_dict,
+        id_col=None,
+        time_col=None,
+        flux_col=None,
+        err_col=None,
+        band_col=None,
+        npartitions=1,
+    ):
+        """Load the sources into an ensemble from a dictionary.
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


src/lsstseries/ensemble.py
```diff
@@ -486,6 +486,70 @@ def from_parquet(
...
+
+        # Load in the source data.
+        self._source = dd.DataFrame.from_dict(source_dict, npartitions=npartitions)
+        self._source = self._source.set_index(self._id_col, drop=True)
+
+        # Generate the object table from the source.
+        self._object = self._generate_object_table()
+
+        # Now synced and clean
+        self._source_dirty = False
```
drewoldag, line 549: Are these set as `True` somewhere prior to this work happening?

drewoldag (approved):
This is pretty tidy. I had just one question, but overall, this looks good.

src/lsstseries/ensemble.py
jeremykubica: They could be. If you have previously loaded data and called sync_tables, they would be set to true. Since this is overwriting the data, we want to make sure they are returned to their initial condition of False.

PR #92, opened 2023-04-24
Structure Function refactor

drewoldag:


src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
drewoldag, line 1: The data manipulation and math was taken directly from the original `_sf2_single` function in `structurefunction2.py`. Looking farther down in this PR, you'll that function has been cut (and pasted here).

src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
```diff
@@ -0,0 +1,126 @@
...
+from lsstseries.analysis.structure_function_argument_containers.structure_function_argument_container import (
+    StructureFunctionArgumentContainer,
+)
+from lsstseries.analysis.structure_function_calculators.structure_function_calculator import (
+    StructureFunctionCalculator,
+)
+
+
+class BasicStructureFunctionCalculator(StructureFunctionCalculator):
+    """The most rudimentary Structure Function calculation method"""
```
drewoldag, line 15: I could use some help from @nevencaplar here (and for traditional documentation) in describing succinctly, what the "basic" structure function is and how it is distinguished from other implementations.

src/lsstseries/analysis/structure_function_calculators/structure_function_calculator.py
```diff
@@ -0,0 +1,93 @@
...
+        self._binning_method = argument_container.bin_method
+        self._bin_count_target = argument_container.bin_count_target
+        self._dts = []
+        return
+
+    def calculate(self):
+        """Abstract method that must be implemented by the child class."""
+        raise (NotImplementedError, "Must be implemented by the child class")
+
+    def _bin_dts(self, dts):
```
drewoldag, line 38: This method was taken from `structurefunction2.py`, and the only modifications here are to allow it to be used as a class method.

src/lsstseries/analysis/structure_function_calculators/structure_function_calculator.py
```diff
@@ -0,0 +1,93 @@
+from typing import List
+
+import numpy as np
+
+from lsstseries.analysis.structure_function_argument_containers.structure_function_argument_container import (
+    StructureFunctionArgumentContainer,
+)
+
+
+class StructureFunctionCalculator:
```
drewoldag, line 10: If we make this a subclass of python's `AbstractBaseClass` then we could use the `@abstractmethod` decorator on the `calculate` method. And I _think_ on the `name_id` and `expected_argument_container` static methods. Doing so would require subclasses of `StructureFunctionCalculator` to implement those methods. If the subclasses failed to implement those methods, then the interpreter would prevent new instances of those classes from being created.

As it stands now, we only raise `NotImplementedError`s which are nice, but don't really enforce anything.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -1,17 +1,14 @@
...
-from scipy.stats import binned_statistic
 
+from lsstseries.analysis.structure_function_calculators import SF_METHODS
 
-def calc_sf2(
-    lc_id, time, flux, err, band, bins=None, band_to_calc=None, combine=False, method="size", sthresh=100
-):
-    """Compute structure function squared on one or many bands
+
+def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", argument_container=None):
```
drewoldag, line 7: This line and the removed line 7, show how the function signature has changed.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -53,9 +39,32 @@ def calc_sf2(
     executed on all available bands in `band`.
     """
 
+    if argument_container is None:
```
drewoldag, line 42: We instantiate a new argument container of the correct type (based on the value of `sf_method`).

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -53,9 +39,32 @@ def calc_sf2(
...
+    if lc_id is None:
+        lc_id = argument_container.lc_id
+
+    if sf_method == "basic":
+        sf_method = argument_container.sf_method
+
+    # Check to make sure that the type of argument container matches what is
+    # required by the structure function calculator method.
+    if type(argument_container) is not SF_METHODS[argument_container.sf_method].expected_argument_container():
+        raise TypeError("Argument container does not match Structure Function calculator method")
```
drewoldag, line 62: If we created the `argument_container` object, we know it will be of the correct type, but it one was passed in, we need to fail fast if it does not match the needs of the Structure Function calculation method.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -112,10 +121,12 @@ def calc_sf2(
             fluxes_2d = [fluxes[mask] for mask in id_masks]
             errors_2d = [errors[mask] for mask in id_masks]
 
-            res = _sf2_single(
-                times_2d, fluxes_2d, errors_2d, bins=bins, combine=combine, method=method, sthresh=sthresh
+            sf_method = SF_METHODS[argument_container.sf_method](
+                times_2d, fluxes_2d, errors_2d, argument_container
```
drewoldag, line 125: Here's where we make use of the `SF_METHODS` dictionary, to create an instance of the structure function calculator object.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -124,188 +135,11 @@ def calc_sf2(
...
         idstack = ["combined"] * len(np.hstack(ids))
     else:
         idstack = np.hstack(ids)
     sf2_df = pd.DataFrame(
         {"lc_id": idstack, "band": np.hstack(bands), "dt": np.hstack(dts), "sf2": np.hstack(sf2s)}
     )
     return sf2_df
-
-
-def _sf2_single(times, fluxes, errors, bins=None, combine=False, method="size", sthresh=100):
```
drewoldag, line 137: The rest of the functions in this file have been moved to either `BasicStructureFunctionCalculator` or `StructureFunctionCalculator`

tests/lsstseries_tests/structure_function_calculators/test_structure_function_calculator.py
drewoldag, line 1: Note about tests: None of the expected values have been updated relative to the original implementation of the tests. The only substantial updates were changes to the order of the input arguments as well as creating and providing an argument_container when we were testing parameters that are no longer available as primary arguments - for example `band_to_calc`.

 The tests in this file have been moved because they belong in a logically different place than they were before.

tests/lsstseries_tests/test_analysis.py
```diff
@@ -95,59 +97,12 @@ def test_sf2_timeseries_with_all_none_timestamps():
...
     timeseries.meta["id"] = 1
     test_series = timeseries.from_dict(data_dict=test_dict)
-    res = test_series.sf2(sthresh=100)
+    res = test_series.sf2()
 
     assert res["dt"][0] == pytest.approx(4.0, rel=0.001)
     assert res["sf2"][0] == pytest.approx(0.005365, rel=0.001)
 
 
-def test_dt_bins():
```
drewoldag, line 104: This and the following test were moved to a new file.

src/lsstseries/analysis/structure_function_argument_containers/structure_function_argument_container.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Notes:
+        It may be necessary to extend this dataclass to support new Structure
+        Function calculation methods. When doing so, all new properties *must*
+        have a default value defined.
+    """
+
+    band: List[str] = field(default=None)
+    lc_id: Union[int, List[int]] = field(default=None)
+    sf_method: str = field(default="basic")
```
delucchi-cmu, line 51: I think you can use a more succinct default value specification, for scalar types at least:

```suggestion
    sf_method: str = "basic"
```

src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
```diff
@@ -0,0 +1,126 @@
+from typing import List
+
+import numpy as np
+from scipy.stats import binned_statistic
+
+from lsstseries.analysis.structure_function_argument_containers.structure_function_argument_container import (
```
delucchi-cmu, line 6: Yikes! That's quite the import!

I'm wondering a different directory structure would be helpful, both for reducing the path lengths and keeping config and implementation closer:

```
_ analysis
  | _ structure_function
     | _ base
     | _ bauer2009
     |  | _ README ## includes more information on formula, citation, etc
     |  | _ container.py
     |  | _ calculator.py
     | _ kozlowski16
     | _ sf2
        | _ README 
        | _ etc.
```

src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
```diff
@@ -0,0 +1,126 @@
...
+                    )
+                    sfs_all.append(sfs)
+                    t_all.append((bin_edgs[0:-1] + bin_edgs[1:]) / 2)
+                else:
+                    sfs_all.append(np.array([]))
+                    t_all.append(np.array([]))
+            return t_all, sfs_all
+
+    @staticmethod
+    def name_id():
```
delucchi-cmu, line 103: Pylint should be ok with inherited methods not having docstrings (and I think it would get more angry about the repeated lines this will generate).

src/lsstseries/analysis/structure_function_calculators/structure_function_calculator.py
delucchi-cmu: +1 to abstract methods.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -1,17 +1,14 @@
...
 
+from lsstseries.analysis.structure_function_calculators import SF_METHODS
 
-def calc_sf2(
-    lc_id, time, flux, err, band, bins=None, band_to_calc=None, combine=False, method="size", sthresh=100
-):
-    """Compute structure function squared on one or many bands
+
+def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", argument_container=None):
+    """_summary_
```
delucchi-cmu, line 8: What's with the new empty summary thing?

src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
nevencaplar: I would avoid the use of the word `rudimentary`. I would include a reference to Equation 12 from https://arxiv.org/abs/1604.05858. I would say something like ``SF calculation method that calculates excess variance directly as a variance of observations with observational errors subtracted.

drewoldag: Thanks - updated!

src/lsstseries/analysis/structurefunction2.py
drewoldag: Gross oversight, fixed!

docs/notebooks/test.ipynb
dougbrn, line 1: you probably are already on top of this, but let's make sure this gets a descriptive name, and maybe less obviously be sure to add this to the toctree in the notebooks rst file

drewoldag: Yes! Great call, I'll polish this notebook, update the name, and get it included in the toctree.

src/lsstseries/analysis/structure_function_calculators/structure_function_calculator.py
drewoldag: Added ABC and @abstractmethod annotations.

src/lsstseries/analysis/structure_function_calculators/basic_structure_function_calculator.py
drewoldag: I've rearranged the directory structure a bit. In some cases it made the import lines shorter, in others it resulted in a few extra characters. 

However, I believe that this will result in more organized code going forward, Thanks for calling this out!

docs/notebooks/test.ipynb
drewoldag: Ok, added more notes to the jupyter notebook and included it in the toctree.

dougbrn (approved):
This looks good to me, happy to approve. Curious to hear if @nevencaplar or @delucchi-cmu have any further comments on this.

drewoldag:
I'll wait until @nevencaplar and @delucchi-cmu sign off before clicking the merge button.

src/lsstseries/analysis/structure_function/basic/calculator.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class BasicStructureFunctionCalculator(StructureFunctionCalculator):
+    """
+    SF calculation method that calculates excess variance directly as a
+    variance of observations with observational errors subtracted.
+    For reference, please see Equation 12 in https://arxiv.org/abs/1604.05858
+    """
+
+    def __init__(
```
delucchi-cmu, line 24: I *think* that if all you're doing is calling super(), you don't need to define init in a subclass. Not doing so would reduce the amount of code expected in the calculator subclasses.

delucchi-cmu (approved):

Issue #88, opened 2023-04-19
LSDB/HiPSCat support


dougbrn:
An action towards #67 is adding support for the [HiPSCat format](https://github.com/astronomy-commons/hipscat), which will be one of the outputs of LSDB.

Support for this involves two main items:
* Implementing a custom loader function that uses a hiPSCat tool to read through their hierarchical directory structure
* Re-indexing the output to be on object id, instead of the custom HiPSCat index.

Other items to be mindful of:
* Ensuring some provenance column is made/kept, to be able to disentangle each survey from the ensemble if needed
* Adding the option to keep the hipscat index as a column, for possible backwards compatibility with LSDB
* Eventually, adding a loader function for the LSDB dataframe directly, avoiding a write to disk.

dougbrn:
#93 Addresses the basics of loading a hive-like directory structure. A future PR will address more hipscat-specific integration, we most likely will want to use LSDB code to facilitate initial joins and are therefore a bit blocked until that functionality makes its way into the main branch of LSDB

nevencaplar:
Closing this, while #339 aims to better capture the needed items.

PR #93, opened 2023-04-25
Hipscat support

dougbrn:
Addresses #88 by adding support for loading from a hive-like directory structure (used by LSDB/hipscat) and adds a provenance column to the source table. I've tested this on epyc with the ZTF dataset, but it doesn't look like there's another hipscat-formatted time-domain survey ready to go. Because of that, I'm not sure if the Ensemble.from_hipscat function as implemented here actually satisfies #88, I am thinking it may make sense to hold this PR open until panstarrs is ready, so that I can test it and update as needed.


src/lsstseries/ensemble.py
```diff
@@ -391,6 +408,69 @@ def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, **k
         else:
             return batch
 
+    def from_hipscat(
```
jeremykubica, line 411: I'm not sure there is anything here that is actually specific to hipscat (except possible the separate source and object directories). We could call this something like "from_parquet_directories" and make it more general.

src/lsstseries/ensemble.py
```diff
@@ -391,6 +408,69 @@ def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, **k
...
+        source_files = glob.glob(os.path.join(source_path, "**", "*.parquet"), recursive=True)
+
+        if object_subdir is not None:
+            object_path = os.path.join(dir, object_subdir)
+            object_files = glob.glob(os.path.join(object_path, "**", "*.parquet"), recursive=True)
+        else:
+            object_files = None
+
+        return self.from_parquet(
+            source_files,
```
jeremykubica, line 464: Should you pass in provenance information here? A natural option would be something like `source_subdir` or adding a provenance parameter to the `from_hipscat` call.

src/lsstseries/ensemble.py
```diff
@@ -453,17 +540,35 @@ def from_parquet(
...
             self._band_col = band_col
+        if provenance_col is not None:
+            self._provenance_col = provenance_col
 
         if additional_cols:
             columns = None  # None will prompt read_parquet to read in all cols
         else:
-            columns = [self._time_col, self._flux_col, self._err_col, self._band_col]
+            if self._provenance_col is not None:
+                columns = [
```
jeremykubica, line 550: This if-else approach is fine for the current set of parameters, but if you are going to include more optional columns then a better approach would be:

```
columns = [self._time_col, self._flux_col, self._err_col, self._band_col]
if self._provenance_col is not None:
    columns.append(self._provenance_col)
```

src/lsstseries/ensemble.py
```diff
@@ -475,7 +580,14 @@ def from_parquet(
...
             # Generate an object table from the source table, then merge
             generated = self._generate_object_table()
-            self._object = file_table.merge(generated, how="right", on=[self._id_col])
+            self._object = file_table.merge(
+                generated, how="right", on=[self._id_col], suffixes=("_original", None)
+            )
+
+            # Drop overwritten columns
+            overwritten_cols = [col for col in list(self._object.columns) if "_original" in col]
+            self._object = self._object.drop(overwritten_cols, axis=1)
```
jeremykubica, line 589: I think we are repeating a lot of the logic from _sync_tables here. Would it be possible to replace this entire block with something like:

```
if object_file:  # read from parquet files
    self._object = self._generate_object_table()
self._sync_tables()
```

src/lsstseries/ensemble.py
```diff
@@ -475,7 +580,14 @@ def from_parquet(
 
             # Generate an object table from the source table, then merge
             generated = self._generate_object_table()
-            self._object = file_table.merge(generated, how="right", on=[self._id_col])
+            self._object = file_table.merge(
```
delucchi-cmu, line 583: The id column may not be the right column to join on, when merging two tables, and hipscat tables will likely use the `_hipscat_index` as the pandas dataframe index for performance reasons.

dougbrn: the `read_parquet` call reindexes based on whichever column was assigned to `self._id_col`, and in this instance the join is between the input object files and an object table generated from the source table. So I *think* the id column makes sense here. You're right in that because it isn't the index it will be slower, but I'm not quite sure what we could do to circumvent that.

src/lsstseries/ensemble.py
```diff
@@ -391,6 +408,69 @@ def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, **k
...
+            critical columns, true will, while false will only load the columns
+            containing the critical quantities (id,time,flux,err,band)
+
+        Returns
+        ----------
+        ensemble: `lsstseries.ensemble.Ensemble`
+            The ensemble object with parquet data loaded
+        """
+
+        source_path = os.path.join(dir, source_subdir)
```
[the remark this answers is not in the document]
dougbrn: Thanks for this comment, it has really got me thinking about what’s missing from this current implementation. Sorry for the incoming brain dump.

For one, you’re definitely correct in that we aren’t utilizing the association information provided by hipscat. We don’t in principle care about this for the actual loading of the data in, as we’re just building two separate dataframes with trust that there’s a common id column to join them in later operations, the only immediate join/merge happening is between the object table and an object table built off the source table. Which, short aside, we don’t have to do this, as the main purpose of it right now is to populate count columns (nobs_total, nobs_r, etc.). We can alternatively just trust those columns as provided by the object table. But where we’ll suffer currently is once we have the data in our dataframes, any time we need to do joins, dask will have to figure out the association as we have no information available for it. 

It’s still not that clear to me how we would use the available hipscat information to supply this to dask dataframes, but one solution to this is to utilize dask dataframe divisions. Which, per partition, denote the minimum and maximum values of the index found in each partition (requires a sorted index). This gets at what you’re suggesting, where it would enable joins to know which partitions in source a given index in object would live. We probably shouldn’t use _id_col as the index in this case, as we’d have to sort it (clobbering any smart partitioning scheme). But we could enumerate the id_col, and use that sorted integer index as our lsstseries index. We can generate that index all within lsstseries, but we still wouldn’t be actually utilizing any information from hipscat. I would have liked to use something like the hipscat_index for this, but the counter is unique per-source, where we need something unique per-object, and mapped to all sources (non-unique). Is there something where we can use the association table to generate an index that is then passed along to the object and source tables? 

The fact that the data format is not set by lsdb/hipscat is actually news to me. Support for alternative data formats is something we should do…in the future.

src/lsstseries/ensemble.py
dougbrn: Ah, yes good catch. But we may want to point to provenance from hipscat. Just looking through their provenance.json files, they seem to indicate properties of the data generation, rather than a single label to give the whole set. We could even by default just use the top level dir as a provenance label (would be "ztf_mar16" in this case)

src/lsstseries/ensemble.py
dougbrn: This is a good idea, implemented.

src/lsstseries/ensemble.py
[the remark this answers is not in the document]
dougbrn: I tried to refactor lsstseries to allow the type of association information that hipscat provides (logged in #98), but ended up with something ~2-5x slower, at least at the scale of 1 million lightcurves or less. At present, we don't do a ton of joins between object and source, really only when doing sync operations between the dataframes. It sounds like potentially what we could do, is use the localized join functionality of LSDB once that has made it into the codebase to generate our object and source tables. This would give us a source table with sources co-located by object id, which for our purposes is ideal.

src/lsstseries/ensemble.py
```diff
@@ -575,6 +633,22 @@ def from_parquet(
...
+        nobs_col: list of 'str', optional
+            Identifies which columns contain number of observations for each
+            band, if available in the input object file
+        nobs_tot_col: 'str', optional
+            Identifies which column contains the total number of observations,
+            if available in the input object file
+        provenance_col: 'str', optional
+            Identifies which column contains the provenance information, if
+            None the provenance column is generated.
+        provenance_label: 'str', optional
```
jeremykubica, line 645: As per your other design document, the list of column names is getting unwieldy. Definitely not worth changing as part of this PR, but it would simplify this code a lot in the future.

src/lsstseries/ensemble.py
jeremykubica: Makes sense.

src/lsstseries/ensemble.py
```diff
@@ -603,37 +677,64 @@ def from_parquet(
...
-            self._source = self._source.repartition(partition_size=partition_size)
-
         if object_file:  # read from parquet files
-            # Read in the object file
-            file_table = dd.read_parquet(object_file, index=self._id_col, split_row_groups=True)
+            # Read in the object file(s)
+            self._object = dd.read_parquet(object_file, index=self._id_col, split_row_groups=True)
+
+            # Handle nobs_band columns
+            if nobs_cols is not None:
```
jeremykubica, line 701: Are these going to be in enough files to justify reading them? Alternatively we could just always auto generate them with sync tables. I understand that we can get it for free it is already in the parquet file, I'm just trying to balance that efficiency vs specializing too much of the code.

src/lsstseries/ensemble.py
```diff
@@ -603,37 +677,64 @@ def from_parquet(
...
+                self._nobs_bands = nobs_cols
+            else:
+                # sets empty nobs cols in object
+                unq_filters = np.unique(self._source[self._band_col])
+                self._nobs_bands = [f"nobs_{filt}" for filt in unq_filters]
+                for col in self._nobs_bands:
+                    self._object[col] = np.nan
```
jeremykubica, line 708: It's not clear to me why you are setting all the values to nan. Why not just leave out the column?

src/lsstseries/ensemble.py
```diff
@@ -753,8 +854,13 @@ def _sync_tables(self):
...
             # Sync Object to Source; remove any missing objects from source
-            self._source = self._source.merge(self._object, how="right", on=[self._id_col])
-            self._source = self._source.drop(list(self._object.columns), axis=1)
+            s_cols = self._source.columns
+            self._source = self._source.merge(
+                self._object, how="right", on=[self._id_col], suffixes=(None, "_obj")
+            )
+            cols_to_drop = [col for col in self._source.columns if col not in s_cols]
+            self._source = self._source.drop(cols_to_drop, axis=1)
+            self._source = self._source.persist()  # persist source
```
jeremykubica, line 863: Why persist the tables (here and below)? That forces computation, right? Wouldn't we want to leave the lazy execution?

src/lsstseries/ensemble.py
dougbrn: Agreed.

src/lsstseries/ensemble.py
dougbrn: So far we know they are available in ZTF and Panstarrs, I wasn't able to find them in the LSST DP0.2 schema, but there was at least a column for the total nobs it looked like. So potentially?

src/lsstseries/ensemble.py
dougbrn: The intention here is to define the columns that will house the nobs_{band} information up front. If a user has set the sync_tables flag, then the sync will automatically update these. But if not, then they would still have access to the columns that *will* house that information once a sync has been performed. I'm also thinking this might clear up some of the ambiguity that would come up if a user loads an object table with nobs cols, but doesn't assign them to lsstseries nobs columns. Particularly by showing in the non-sync cases that new columns have been added that will dynamically update, and those aren't the same as the columns with pre-populated values found from the file.

src/lsstseries/ensemble.py
dougbrn: Good question. I settled on adding persist calls here because they ended up being the difference between my testing with the ~1 million lightcurve ZTF set working and not working. The key thing about persist calls, is that while they do submit the task graph to the scheduler, they return a dask futures object not an in-memory object (like a pandas dataframe). So it's still lazy execution in the sense that we aren't requesting any part of the data to be brought into memory.

As for what this is actually fixing in my test, I was running into a lot of shuffle transfer failures without persist calls. And I'm thinking these were due to the task graph becoming too large, with too many complex commands being queued. The persist calls seem to ensure that manageable task graphs are being submitted. As instead of one gigantic task graph generated off of ensemble._source or ensemble._object, we instead package smaller task graphs into futures objects and create new task graphs that operate on those futures... That being said, I could stand to learn a lot more about what is actually going on under the hood, but persist calls seems to be generally recommended and seem to help facilitate complex workflows at scale.

src/lsstseries/ensemble.py
jeremykubica: We can keep it in the code for now, but I'm worried about adding too much column-specific logic.

src/lsstseries/ensemble.py
jeremykubica: Makes sense. Thanks.

src/lsstseries/ensemble.py
jeremykubica: Thanks! That is very helpful.

jeremykubica (approved):

PR #95, opened 2023-04-28
Adding support to allow users to pass `None` for lc_id and/or band

drewoldag:
The vast majority of this PR is organizing argument validation in `structurefunction2.py`. You'll see calls to functions named `_validate_<arg>`. In each of those we'll take a look at the input arguments and possibly the values found in argument_container as well, and take the one with the highest priority. 

Additionally we may do some comparison checks and raise exceptions if the inputs are not well formed. i.e. number of elements in `band` doesn't match number of elements in `flux` arrays.

But to address the title of the PR, this work also allows users:
- Pass nothing at all for `lc_id` (`None` will be the default)
- Pass `None` for `lc_id`
- Pass an array for `lc_id`
- Pass `lc_id` inside of the argument container

and 
- Pass nothing at all for `band` (`None` will be the default)
- Pass `None` for `band`
- Pass an array for `band`
- Pass `band` inside of the argument container

### DIFF
--- a/src/lsstseries/analysis/structurefunction2.py
+++ b/src/lsstseries/analysis/structurefunction2.py
@@ -53,9 +53,13 @@ def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", arg
     # `argument_container` and use the values found there.
     if band is None:
         band = argument_container.band
+    if band is None:  # Still None after assignment?
+        band = np.full(len(flux), "none")
 
     if lc_id is None:
         lc_id = argument_container.lc_id
+    if lc_id is None:  # Still None after assignment?
+        lc_id = np.zeros(len(flux))
 
     if sf_method == "basic":
         sf_method = argument_container.sf_method
@@ -91,7 +95,7 @@ def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", arg
             # with a single `None` element. Otherwise assume the user passed an
             # array of timestamps to be masked with `band_mask`.
             # Note: some or all timestamps could be `None`.
-            if time is None:
+            if time is None or argument_container.ignore_timestamps:
                 times = np.array(None)
             else:
                 times = np.array(time)[band_mask]
--- a/src/lsstseries/analysis/structurefunction2.py
+++ b/src/lsstseries/analysis/structurefunction2.py
@@ -51,10 +51,7 @@ def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", arg
     # `argument_container`. If any of these arguments are provided with
     # non-default values, we'll use those. Otherwise, we'll look inside
     # `argument_container` and use the values found there.
-    if band is None:
-        band = argument_container.band
-    if band is None:  # Still None after assignment?
-        band = np.full(len(flux), "none")
+    band = _validate_band(band, flux, argument_container)
 
     if lc_id is None:
         lc_id = argument_container.lc_id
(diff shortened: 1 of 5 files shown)


src/lsstseries/analysis/structurefunction2.py
```diff
@@ -51,11 +51,12 @@ def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", arg
     # `argument_container`. If any of these arguments are provided with
     # non-default values, we'll use those. Otherwise, we'll look inside
     # `argument_container` and use the values found there.
-    if band is None:
-        band = argument_container.band
+    band = _validate_band(band, flux, argument_container)
 
     if lc_id is None:
         lc_id = argument_container.lc_id
+    if lc_id is None:  # Still None after assignment?
```
nevencaplar, line 58: Use of the question makes this a little bit unclear. If I understand correctly, you are pointing to the person who is reading the code that this will be executed if there is no `lc_id` in the argument_container. I would say that more explicitly, `used if still None after assignment` or `used if not specified in after argument_container`

nevencaplar: Right, I see later you were more explicit in `_validate_band` function. I would apply same explanations (or same level of explanation) here.

drewoldag: Yep, good comment. I committed some code last week so that I would have it handy, but it clearly wasn't polished. The most recent commit has corrected that, and brought that section of the input argument validation checks into sync with the `_validate_band` function.

tests/lsstseries_tests/test_analysis.py
```diff
@@ -239,3 +239,308 @@ def test_sf2_base_case_error_as_none():
...
+    """
+    Base test case accessing calc_sf2 directly. Pass time as None and identical
+    flux values, but nothing else.
+    Does not make use of TimeSeries or Ensemble.
+    """
+    test_y = [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
+
+    res = analysis.calc_sf2(time=None, flux=test_y)
+
+    # assert res["dt"][0] == pytest.approx(3.705, rel=0.001)
```
dougbrn, line 316: here as well

dougbrn (approved):
This looks good to me, just a few minor comments

tests/lsstseries_tests/test_analysis.py
drewoldag: Good catch, thank you!

tests/lsstseries_tests/test_analysis.py
```diff
@@ -239,3 +239,308 @@ def test_sf2_base_case_error_as_none():
...
+    Does not make use of TimeSeries or Ensemble.
+    """
+    test_y = [0.11, 0.23, 0.45, 0.01, 0.67, 0.32, 0.88, 0.2]
+
+    res = analysis.calc_sf2(
+        time=None,
+        flux=test_y,
+    )
+
+    # assert res["dt"][0] == pytest.approx(3.705, rel=0.001)
```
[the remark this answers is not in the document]
drewoldag: Good call!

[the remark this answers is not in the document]
drewoldag: Fixed

tests/lsstseries_tests/test_analysis.py
drewoldag: Fixed!

Issue #94, opened 2023-04-26
Add a demo function of the binning function


jeremykubica:
Add examples to some combination of the documentation and notebooks.

PR #96, opened 2023-04-28
Add an example notebook for binning

jeremykubica:
Addresses issue #94


docs/notebooks/binning_slowly_changing_sources.ipynb
```diff
@@ -0,0 +1,369 @@
+{
```
nevencaplar, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    _ = plt.hist(ens._source["midPointTai"].compute().tolist(), 1000)
</pre><p>All plots need to have labels, e.g., on first plot the x-axis should be either MJD, or time, or days, or midPointTai, and y-axis should be number of objects of N</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lsstseries/pull/96/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

nevencaplar, line 1: <p>I wonder if sentence The <code>offset</code> parameter allows the user to indicate where the division point should be between days'' can be clearer. I think the confusion arrises because we use a word day do mean both 24 hour period and the bright part of the that period (i.e., opposite to night). In this case these meanings are hard to distinguish. Something more like - The <code>offset</code> parameter allows the user to indicate where the division between different nights/observing blocks should be. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lsstseries/pull/96/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

nevencaplar, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    ens.bin_sources(time_window=7.0, offset=0.0)
</pre><p>What happens if you bin with non-integer value? We basically do not want that, right? Perhaps this should be forced to be an int?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lsstseries/pull/96/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

nevencaplar, line 1: <p>More explanation needed here. This jumps somewhat and assumes that user know about compute and lazy evaluation in ensamble objects. If they do not know, they could wonder where did this come from.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lsstseries/pull/96/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

jeremykubica: Done.

jeremykubica: I like your phrasing better and used that.

jeremykubica: This should still work with a non-integer value for `time_window`. The concept of an `offset` is a bit weird at less than a 24 hour binning. In this case `offset` still shifts the bins but it does not need to be aligned with the middle of the daylight hours.

jeremykubica: I added more.

nevencaplar: Hm... Should we maybe make a warning if you are trying do time_window with floats? I am worried about a case where algorithm silently make results that users may not expect.

jeremykubica: I don't think that we need a warning. The algorithm will still bin correctly. I imagine 1 hour binning will likely be a valid (and useful) case.

nevencaplar (approved):
Looks good to me

PR #97, opened 2023-04-28
Add ability for source binning to save the counts

jeremykubica:


dougbrn (approved):
Nice, this looks good to me

PR #99, opened 2023-05-01
Updating copier to v1.3.3 and removing isort from pre-commit

drewoldag:
The main driver here was removing isort from pre-commit. It could have been done manually, but I noticed that the copier version was getting a little behind, so I took the opportunity to update it at the same time.


dougbrn (approved):
Looks good to me

PR #100, opened 2023-05-02
Implementing IQR structure function calculator

drewoldag:


src/lsstseries/analysis/structure_function/base_calculator.py
```diff
@@ -14,9 +14,9 @@ class StructureFunctionCalculator(ABC):
 
     def __init__(
         self,
-        time: List[float],
-        flux: List[float],
-        err: List[float],
+        time: List[List[float]],
+        flux: List[List[float]],
+        err: List[List[float]],
```
drewoldag, line 19: I just wanted to be more explicit here, since we do expect that these inputs will be 2d arrays.

src/lsstseries/analysis/structure_function/basic/calculator.py
```diff
@@ -16,9 +16,9 @@ class BasicStructureFunctionCalculator(StructureFunctionCalculator):
 
     def __init__(
         self,
-        time: List[float],
-        flux: List[float],
-        err: List[float],
+        time: List[List[float]],
+        flux: List[List[float]],
+        err: List[List[float]],
```
drewoldag, line 21: Same comment as in `base_calculator.py`.

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+    def __init__(
+        self,
+        time: List[List[float]],
+        flux: List[List[float]],
+        err: List[List[float]],
+        argument_container: StructureFunctionArgumentContainer,
+    ):
+        super().__init__(time, flux, err, argument_container)
+
+    def calculate(self):
```
drewoldag, line 32: In a follow up PR, I would like to abstract the majority of the work here in `calculate` into the parent class. It seems like each calculation method is going to require very similar data manipulation at this step, and I believe that the majority of the code in lines 33-57 could be one parent class method, and then lines 60-73 could be another parent class method that would accept an input argument that defined what statistic to use in `binned_statistic`.

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+                    # binning
+                    self._bin_dts(self._dts[lc_idx])
+                    sfs, bin_edgs, _ = binned_statistic(
+                        self._dts[lc_idx],
+                        all_d_fluxes[lc_idx],
+                        statistic=self.calculate_iqr_sf2_statistic,
+                        bins=self._bins,
+                    )
+                    sfs_all.append(sfs)
+                    t_all.append((bin_edgs[0:-1] + bin_edgs[1:]) / 2)
```
nevencaplar, line 91: Is this how the initial implementation was as well? Mean of the bin edges? I think that the mean of the times that go into the bin would be preferable.

drewoldag: Yeah, this was copied over, but I can refactor to make that change and return mean of the delta_time values in the bin.

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+from lsstseries.analysis.structure_function.base_argument_container import StructureFunctionArgumentContainer
+from lsstseries.analysis.structure_function.base_calculator import StructureFunctionCalculator
+
+# For details see Kozlowski 16 Equation 10: https://arxiv.org/abs/1604.05858
+CONVERSION_TO_SIGMA = 0.741
+
+
+class IqrStructureFunctionCalculator(StructureFunctionCalculator):
+    """This class implements the structure function calculation described in
+    Equation 10 of Kozlowski 16: https://arxiv.org/abs/1604.05858
```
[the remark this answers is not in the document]
drewoldag: The MacLeod 2012 equation includes the `1/sqrt(N-1)`, but this SF calculation method does not. This method really is just 0.741 * IQR, which is what's shown in this paper as eqn 10. https://arxiv.org/pdf/1604.05858.pdf.

I'm implementing the MacLeod 2012 method in PR #101, specifically (https://github.com/lincc-frameworks/lsstseries/pull/101/files#diff-1624a07efba9682b4ead1cc37b2d14dc00d63d3f98f26c711b028b13eeb50ebc)

src/lsstseries/analysis/structure_function/base_calculator.py
jeremykubica: Would these be more efficient as numpy arrays? (Not necessarily in scope of this change).

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+
+            # build stack of times and fluxes
+            # `self._dts` and `all_d_fluxes` will have shape = (num_lightcurves, N)
+            self._dts.append(d_times)
+            all_d_fluxes.append(d_fluxes)
+
+        # combining treats all lightcurves as one when calculating the structure function
+        if self._argument_container.combine and len(self._time) > 1:
+            self._dts = np.hstack(np.array(self._dts, dtype="object"))
+            all_d_fluxes = np.hstack(np.array(all_d_fluxes, dtype="object"))
```
jeremykubica, line 62: Just checking that I understand: This concatenates all of the light curve deltas (deltas in both time and flux). But the zero points of each of these light curves will be different. Would it make more sense to concatenate before computing `dt_matrix` and `df_matrix` so the light curves are all baselined to the same zero value?

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+            t_mask = np.isnan(lc_times)
+            f_mask = np.isnan(lc_fluxes)
+            lc_mask = np.logical_or(t_mask, f_mask)
+
+            lc_times = lc_times[~lc_mask]
+            lc_fluxes = lc_fluxes[~lc_mask]
+
+            # compute difference of times
+            dt_matrix = lc_times.reshape((1, lc_times.size)) - lc_times.reshape((lc_times.size, 1))
+            d_times = dt_matrix[dt_matrix > 0].flatten()
```
jeremykubica, line 48: In the paper you link, they define delta_t as abs(t_i - t_j). And then use the different pairings and a scale factor equal to the number of pairings. I think this is *mostly* (but not quite exactly) the same because you are throwing out approximately half the data (all the negative), which will effect the list of pairs and the number of pairings.

I'm not sure about the case of dt_matrix = 0 though. From a quick scan through the paper I'm not sure if the t_0 - t_0 is included in the pairings. If so that would be a data point for the IQR that you are dropping and could shift the boundaries a bit.

There is also the implicit assumption that there is only one observation at the first time. The pairings will be off if we have multiple observations at t_0 because they will all be filtered.

jeremykubica (commented):
The code largely looks good. However I have some questions (for my own understanding) about the underlying math.

src/lsstseries/analysis/structure_function/iqr/calculator.py
nevencaplar: Keep it as such, we may want to have separate issue in the future.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: This is good call out, they actually are passed in as 2d numpy arrays. And it prompts me to look up how to do type hints for numpy arrays. Looks like this is at least one approach: https://numpy.org/doc/stable/reference/typing.html#module-numpy.typing

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Hmm, this is a good question. I'll have to ask Neven about this one. However, I can see that this code process is similar (identical) to what is show in this notebook: https://github.com/ywx649999311/structure_function/blob/master/notebooks/SF.ipynb that Neven shared earlier. 

But that being said, I'll spend some time with Neven to make sure that we're correctly interpreting what's in the paper. 

As another point of reference, we're using this paper as the source for several other SF methods https://arxiv.org/pdf/1401.1785.pdf and it's a little more clear about what constitutes a delta_t. 

"where ∆mij is the magnitude difference between the ith and jth observations (i < j) whose time lag (tj − ti) falls in the bin ∆t and with measurement uncertainties of σi and σj respectively"

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Understood. I'll leave it as is for now.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: Done

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
+from typing import List
+
+import numpy as np
+from scipy.stats import binned_statistic
+
+from lsstseries.analysis.structure_function.base_argument_container import StructureFunctionArgumentContainer
+from lsstseries.analysis.structure_function.base_calculator import StructureFunctionCalculator
+
+# For details see Kozlowski 16 Equation 10: https://arxiv.org/abs/1604.05858
```
drewoldag, line 9: Change this to Kozlowski 2016

src/lsstseries/analysis/structure_function/iqr/calculator.py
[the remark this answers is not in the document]
drewoldag: MacLeod et al. 2012, 2012ApJ...753..106M [https://arxiv.org/abs/1112.0679]

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Just finished up a chat with Neven on this one. The approach here is considered correct. We're considering the time difference between observations at t_i and t_j where i < j. And you're correct, if there are multiple observations at the same time, we would not include the delta_magnitudes between those observations. 

Depending on the granularity of the timestamps, it might be a very rare case when two observations would happen at the same time.  If the finest timescale is a day, then this might be more common.

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Also chatted with Neven about this - here we want to combine the delta_t and delta_magnitudes, not the observation timestamps and observed magnitudes. So in this case, we're doing what is expected.

src/lsstseries/analysis/structure_function/iqr/calculator.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Parameters
+        ----------
+        input : `np.ndarray` (N,)
+            The delta flux values that correspond to a given delta time bin.
+
+        Returns
+        -------
+        float
+            Result of calculating Equation 10 of Kozlowski 16:
```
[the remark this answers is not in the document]
drewoldag: Done

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Created this issue to update this shortly.

src/lsstseries/analysis/structure_function/iqr/calculator.py
drewoldag: Created an issue for this here: https://github.com/lincc-frameworks/lsstseries/issues/103

nevencaplar (approved):
Science part looks good to me

jeremykubica (approved):

PR #105, opened 2023-05-05
Correct pre-commit sphinx configuration, add __init__ files.

drewoldag:
Updates to the sphinx configuration for the pre-commit hook weren't correctly applied when I updated the copier template last time. This PR fixes that. 

I also noticed that the new subclasses of Structure Function calculator class weren't being picked up by AutoAPI - I needed to add a __init__.py file to the directories so they would identified and included.


dougbrn (approved):
Looks good to me

PR #108, opened 2023-05-11
Return `mean(delta_t values in bin)` as `dts` for Structure Function

drewoldag:
In the main branch after we have established the bins for the time differences, we return the bin centers to the user as the parameter `dts`. However, in practice this value can be misleading, and it is more useful to return the mean of the values in a given bin as the parameter `dts`. 

This PR represents the work to gather those statistics for the bins. There were several unit tests that needed to be updated as well, but because we were using small, contrived examples, it was not difficult to verify that the new `dts` values were correct.

### DIFF
--- a/src/lsstseries/analysis/structure_function/base_calculator.py
+++ b/src/lsstseries/analysis/structure_function/base_calculator.py
@@ -146,8 +146,8 @@ class StructureFunctionCalculator(ABC):
                 )
 
             # return the mean delta_time values for each bin
-            # bin_means, _, _ = binned_statistic(self._dts, self._dts, statistic="mean", bins=self._bins)
-            return [(bin_edgs[0:-1] + bin_edgs[1:]) / 2], [sfs]
+            bin_means, _, _ = binned_statistic(self._dts, self._dts, statistic="mean", bins=self._bins)
+            return [bin_means], [sfs]
 
         # Not combining calculates structure function for each light curve independently
         else:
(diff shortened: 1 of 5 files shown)


src/lsstseries/analysis/structure_function/base_calculator.py
```diff
@@ -146,8 +146,8 @@ def _calculate_binned_statistics(self, sample_values=None, statistic_to_apply="m
                 )
 
             # return the mean delta_time values for each bin
-            # bin_means, _, _ = binned_statistic(self._dts, self._dts, statistic="mean", bins=self._bins)
-            return [(bin_edgs[0:-1] + bin_edgs[1:]) / 2], [sfs]
+            bin_means, _, _ = binned_statistic(self._dts, self._dts, statistic="mean", bins=self._bins)
+            return [bin_means], [sfs]
```
smcguire-cmu, line 150: Should the docstring for this function be updated to specify that it returns the `bin_means` now? I see it still mentions the edges (but that might be for another reason)

smcguire-cmu (commented):
It looks like a simple enough change, and I don't think there's anything wrong with the code written here, but I don't think I really know enough about how it fits in with the rest of the codebase to properly review it.

smcguire-cmu (commented):
It looks like a simple enough change, and I don't think there's anything wrong with the code written here, but I don't think I really know enough about how it fits in with the rest of the codebase to properly review it.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: Good call out, I'll update the docstring.

drewoldag: Updated

smcguire-cmu (approved):
Looks good to me, but might want to wait until approval from Doug or Neven. (Unless you've already talked about it)

nevencaplar (approved):
Looks good to me

PR #113, opened 2023-05-12
Fixes bug that causes Structure Function to ignore user provided time bins.

drewoldag:
In the non-combined mode of Structure Function calculations, we would ignore user provided bins in the `argument_container` and instead always calculate the bins. This PR fixes that bug and introduces regression unit tests.

### DIFF
--- a/src/lsstseries/analysis/structure_function/base_calculator.py
+++ b/src/lsstseries/analysis/structure_function/base_calculator.py
@@ -168,7 +168,7 @@ class StructureFunctionCalculator(ABC):
             # structure function at specific dt
             # the line below will throw error if the bins are not covering the whole range
             try:
-                sfs, bin_edgs, _ = binned_statistic(
+                sfs, _, _ = binned_statistic(
                     self._dts, sample_values, statistic=statistic_to_apply, bins=self._bins
                 )
             except AttributeError:
@@ -189,10 +189,12 @@ class StructureFunctionCalculator(ABC):
                 if len(self._dts[lc_idx]) > 1:
                     # bin the delta_time values, and evaluate the `statistic_to_apply`
                     # for the delta_flux values in each bin.
-                    self._bin_dts(self._dts[lc_idx])
+
+                    if self._bins is None:
+                        self._bin_dts(self._dts[lc_idx])
 
                     try:
-                        sfs, bin_edgs, _ = binned_statistic(
+                        sfs, _, _ = binned_statistic(
                             self._dts[lc_idx],
                             sample_values[lc_idx],
                             statistic=statistic_to_apply,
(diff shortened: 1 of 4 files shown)


src/lsstseries/analysis/structure_function/base_calculator.py
```diff
@@ -189,10 +189,12 @@ def _calculate_binned_statistics(self, sample_values=None, statistic_to_apply="m
                 if len(self._dts[lc_idx]) > 1:
                     # bin the delta_time values, and evaluate the `statistic_to_apply`
                     # for the delta_flux values in each bin.
-                    self._bin_dts(self._dts[lc_idx])
+
+                    if self._bins is None:
```
drewoldag, line 193: This is the line that fixes the bug.

src/lsstseries/analysis/structure_function/base_calculator.py
```diff
@@ -189,10 +189,12 @@ def _calculate_binned_statistics(self, sample_values=None, statistic_to_apply="m
...
+                        self._bin_dts(self._dts[lc_idx])
 
                     try:
-                        sfs, bin_edgs, _ = binned_statistic(
+                        sfs, _, _ = binned_statistic(
```
drewoldag, line 197: `bin_edges` was unused. Same story on line 171

src/lsstseries/analysis/structure_function/base_calculator.py
cchris28: This is the bug fix, and I see you removed some unused lines.
Do the three unit tests you added test this bug? It isn't obvious to me so I ask for confirmation.

tests/lsstseries_tests/test_analysis.py
```diff
@@ -557,3 +557,88 @@ def test_sf2_base_case_macleod_2012():
...
+    or Ensemble.
+    """
+    lc_id = [1, 1, 1, 1, 1, 1, 1, 1]
+    test_t = [1.11, 2.23, 3.45, 4.01, 5.67, 6.32, 7.88, 8.2]
+    test_y = [0.11, 0.23, 0.45, 0.01, 0.67, 0.32, 0.88, 0.2]
+    test_yerr = [0.1, 0.023, 0.045, 0.1, 0.067, 0.032, 0.8, 0.02]
+    test_band = np.array(["r"] * len(test_y))
+
+    arg_container = StructureFunctionArgumentContainer()
+    arg_container.bins = [0.0, 3.1, 9.0]
```
drewoldag, line 635: This test specifies a particular set of bin edges, and ensures they are used.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: The last unit test in test_analysis is specifically protecting this change from regressions. The other tests are just for increased code coverage.

drewoldag: I added a comment on the line in `test_analysis` that is actually doing the work :)

cchris28 (approved):
lgtm

Issue #61, opened 2023-04-05
Add possibility in structure function that objects have the same weight


nevencaplar:
When working with an ensemble (many lightcurves) we do not want that one or few lightcurves over-contribute to the estimate of the structure function.

Add a possibility to sample for each lightcurve number of pairs that correspond to the number of pairs you are getting from the lightcurve with the smallest number of points.

For example; two lightcurves, one with N_1 = 10, N_2=100, where N_i is the number of points in each lightcurve. The first one gives 10*9/2 = 45 pairs, the second gives 100*99/2 = 4950 pairs. Sample 45 pairs from each lightcurve. I think this is better than sampling 10 points from each curve, even though it is potentially more expensive.

drewoldag:
Notes for my future self. Numpy's np.random.choice(np.array, N, replace=False) seems like a useful tool for this work.

drewoldag:
Notes from conversations: 

- Operating per band. Each band is essentially it's own object, so for instance N_1 with 100 r-band obs and 10 g-band obs, should not subsample the r-band.
- We should operate on pairs of values (i.e. the differences). Keep in mind that we'll also want to 'bootstrap' (resample multiple times as in issue #62) so it will be good to keep the original full set of pairs, so that we can quickly resample from it multiple times without having to recalculate it. 
- Keep an eye on memory/cpu usage. Estimating that we could get into the regime of several thousands of observations. Should investigate what the practical limits are.

PR #114, opened 2023-05-12
Consolidate duplicated Structure Function code into `base_calculator.py`.

drewoldag:
I realize that this PR looks nasty. 

Each of the different Structure Function calculator methods had a large block of duplicated code that has now been consolidated into base_calculator. 

Prior to starting on the work to subsample and bootstrap the data (issue #61), it is necessary to consolidate the code to avoid making complicated changed in 5 different places. 

It's good to note that all the unit tests still pass despite this large change, and now, there's 1/5 the amount of code to maintain.


dougbrn (approved):
Looks good to me, less code to maintain is good!

Issue #62, opened 2023-04-05
Estimate error on the measurement of structure function in each delta_t bin


nevencaplar:
The goal is to enable an estimate of the uncertainty on the calculated structure function.

Resample pairs of measurements randomly with repetition M times (I recommend 100 times; possibly add exact value of M as an option).
If you are in ``object have the same weight mode’’, select new N measurements from lightcurve that have more measurements each time.
Calculate the median of these 100 measurements and 16 to 83 quantile - report that as 1-sigma errors of the measurement (allow for the possibility to report different upper and lower errors).

drewoldag:
Report 1_sigma as (83-16)/2 
Provide option for user to output lower_error as median-16 and upper_error as 83-median.

Issue #68, opened 2023-04-05
Tutorial for structure function


nevencaplar:
Create a tutorial notebook which covers functionality implemented in [59](https://github.com/lincc-frameworks/lsstseries/issues/59), [60](https://github.com/lincc-frameworks/lsstseries/issues/60), [61](https://github.com/lincc-frameworks/lsstseries/issues/61), [62](https://github.com/lincc-frameworks/lsstseries/issues/62)

drewoldag:
Be sure to include information about how to work with StructureFunctionLightCurve objects, both the primary and derived data.

PR #115, opened 2023-05-18
Equally weight lightcurves

drewoldag:
This PR introduces the StructureFunctionLightCurve object to be used instead of three arrays that need to be kept in sync.

This makes it much easier to implement the features requested in #61 and #62. 

This PR also represents the work to subsample light curves. The user can specify whether they would like a specific number of differences sampled or default to sampling equal to the lightcurve with the fewest differences. 

Additionally this PR implements bootstraping to allow the user to repeat the Structure Function calculation N times and collect median and error metrics from those. By default the 1 sigma error will be reported, but the user can optionally request that upper and lower error values be reported. The user also has the ability to specify the upper and lower quantiles used to calculate the error values.


src/lsstseries/analysis/structure_function/light_curve.py
```diff
@@ -0,0 +1,131 @@
+from typing import Optional
+import numpy as np
+
+MIN_OBSERVATIONS_REQUIRED_FOR_SF = 2
```
nevencaplar, line 4: I think this should be at least 3. If we have 2 datapoints, that is one difference of times, and that has no meaning when we measure variance of one data point.

src/lsstseries/analysis/structure_function/light_curve.py
```diff
@@ -0,0 +1,131 @@
...
+            allows for repeatable random samples to be selected. By default None
+
+        Raises
+        ------
+        ValueError
+            If samples are requested than are present in the light curve raise
+            ValueError.
+
+        Notes
+        -----
```
nevencaplar, line 109: Empty notes section?

src/lsstseries/analysis/structure_function/light_curve.py
```diff
@@ -0,0 +1,131 @@
...
+        differences.
+
+        Parameters
+        ----------
+        number_of_samples : int
+            Defines the number of samples to be randomly selected from the total
+            number of difference values.
+        random_generator: np.random.Generator, optional
+            A Numpy random.Generator to sample the lightcurve difference. This
+            allows for repeatable random samples to be selected. By default None
```
nevencaplar, line 100: End sentence with '.' to be consistent.

src/lsstseries/analysis/structurefunction2.py
```diff
@@ -81,29 +87,82 @@ def calc_sf2(time, flux, err=None, band=None, lc_id=None, sf_method="basic", arg
...
+            lightcurves = []
+            for mask in id_masks:
+                sf_lc = StructureFunctionLightCurve(
+                    times=times[mask], fluxes=fluxes[mask], errors=errors[mask]
+                )
+                lightcurves.append(sf_lc)
+
+            sf_calculator = SF_METHODS[sf_method](lightcurves, argument_container)
+
+            # aggregated_* has shape [calc_rep(0:arg_container.calc_repetitions)][lc_id(0:num_lightcurves)][bin(0:num_dt_bins)]
```
nevencaplar, line 100: Is this a relic of some previous code that was not deleted?

src/lsstseries/analysis/structure_function/light_curve.py
```diff
@@ -0,0 +1,131 @@
+from typing import Optional
+import numpy as np
+
+MIN_OBSERVATIONS_REQUIRED_FOR_SF = 2
+
+
+class StructureFunctionLightCurve:
```
dougbrn, line 7: Do we envision other analysis functions needing to create a LightCurve object? A lot of this specific code is handling the structurefunction case, but in the event that say, periodograms, needed a PeriodogramLightCurve class, it possibly could make sense to have a LightCurve baseclass established? Not sure

src/lsstseries/analysis/structure_function/base_calculator.py
```diff
@@ -46,74 +47,46 @@ class StructureFunctionCalculator(ABC):
...
-            d_fluxes = df_matrix[dt_matrix > 0].flatten()
-            self._all_d_fluxes.append(d_fluxes)
-
-            # err^2 = sum of squared errors
-            err2_matrix = (
-                lc_errors.reshape((1, lc_errors.size)) ** 2 + lc_errors.reshape((lc_errors.size, 1)) ** 2
-            )
-            err2s = err2_matrix[dt_matrix > 0].flatten()
-            self._sum_error_squared.append(err2s)
+        self._difference_values_per_lightcurve = [lc.number_of_difference_values for lc in self._lightcurves]
```
dougbrn, line 89: I'm struggling a bit to understand what the dividing line is between analysis done in the StructureFunctionLightCurve object is versus analysis done in the StructureFunction code itself. Is it just that the lightcurve class is handling all the data serving tasks (difference calculation, bootstrapping), and then the analysis function itself is using those to calculate the sf values? Particularly from a user perspective, I'd be worried about if I had my own structure function, what if I needed it to do something differently with the data? Am I then trying to modify the StructureFunctionLightCurve? Part of this may just be me needing to read this a bit more, but figured I'd just flag my ignorance here.

src/lsstseries/analysis/structure_function/light_curve.py
drewoldag: Done

src/lsstseries/analysis/structure_function/light_curve.py
drewoldag: Removed

src/lsstseries/analysis/structure_function/light_curve.py
drewoldag: That makes sense, I'll update this.

src/lsstseries/analysis/structure_function/light_curve.py
drewoldag: That is a fair a point. Initially I was envisioning this as just being used by StructureFunction. However, it wouldn't be too hard to abstract this up one level to create a generic Lightcurve base class for that StructureFunctionLightCurve would inherit from. 

I could see the base class looking something like:

```class AnalysisLightCurve:  # name TBD
    def __init__(self, times, fluxes, errors):
        self._times = times
        self._fluxes = fluxes
        self._errors = errors

        self._process_input_data()
   
    def _process_input_data(self):
        # implement each of this in this class as well
        self._check_input_data_size_is_equal()
        self._filter_nans()
        self._check_input_data_length_is_sufficient()
```

src/lsstseries/analysis/structurefunction2.py
drewoldag: This is meant to be a comment indicating the resulting shape of the following two variables. I found it challenging to keep track of what each of the dimensions represented, so I wanted to make a note of it here. However, if it's not intuitive, then I can rework this to make it more clear.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: I think what you're sensing is justified, the line between the data manipulation that happens in SFLightCurve and the calculator can be hazy. The way I was viewing this was that the SFLightCurve object would contain the primary data, as well as the differences. Additionally, it would contain the data validation (NaN's, Lists have same number of elements etc.) 

I included the subset selection routine in the class as well for what I perceived as efficiency reasons. The class could maintain the full set of differences in memory, and we could select a subset of those multiple times without having to recreate anything. That function could have gone into `calc_sf2` as well, but I figured it made more sense in the SFLightCurve class. 

I viewed the calculator as doing the Structure Function calculations on one or more SFLightCurve objects (and optionally doing the "combine" action as well). 

To your point about a user with their own structure function, the goal would be that they would not have to edit or subclass the SFLightCurve class. A user would have access to the primary data in the SFLightCurve via `self._times self._fluxes and self._errors`. They would also have access to the full set of difference values via `self.sample_d_times self.sample_d_fluxes and self.sample_sum_squared_error`. (If the user does not apply subsampling to the light curve, each of those represent the full set of difference values. 

So if a user wanted to do something differently with the data, they would be able to manipulate the primary input data within the `calculate` method of their `StructureFunctionCalculator` subclass, or they could manipulate the differences as they see fit. 

That was a super wordy response on my part. I'm happy to chat this out in person if that's easier, because if you're getting this feeling that this is the wrong abstraction, then I want to hear more about it :)

dougbrn: Thanks for the detailed response Drew, I'm on board with the idea that we have a class for handling details of the lightcurve (e.g. is it valid? Scrub it for NaNs up front, etc.), and I'm leaning more and more into thinking we want a general lightcurve class that does data validation for any arbitrary analysis function. And then inherited classes like SFlightcurve get to do all the extra bits for their science function. 

And I take your point regarding users still having the equivalent amount of information to work with outside of needing to do any modification of these lightcurve classes, that seems reasonable.

My main concern here is just that I want to make sure that any user who wants to adapt/implement their own science code is able to actually build on top of our codebase without excessive confusion. I think documentation is a perfectly good answer to that for where we are right now.

I'll give this another pass today since it's a big PR, but consider this issue I raised addressed :)

nevencaplar:
Looks good to me, happy to merge after Doug's comments get resolved.

src/lsstseries/analysis/structure_function/light_curve.py
drewoldag: I created Issue #122 to track this work.

src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag: Added a comment on issue #68 to be sure to describe working with StructureFunctionLightCurve instances.

dougbrn (approved):
This looks good to me! Thanks for making an issue for the LightCurve class

PR #116, opened 2023-05-23
Colmapper

dougbrn:
Implements the ColumnMapper class in lsstseries.utils


src/lsstseries/utils/column_mapper/column_mapper.py
```diff
@@ -0,0 +1,200 @@
...
+            Identifies which column contains the Object IDs
+        time_col: 'str', optional
+            Identifies which column contains the time information
+        flux_col: 'str', optional
+            Identifies which column contains the flux/magnitude information
+        err_col: 'str', optional
+            Identifies which column contains the flux/mag error information
+        band_col: 'str', optional
+            Identifies which column contains the band information
+        nobs_col: list of 'str', optional
```
drewoldag, line 29: Update to be `nobs_band_cols`.

src/lsstseries/utils/column_mapper/column_mapper.py
```diff
@@ -0,0 +1,200 @@
...
+            Identifies which columns contain number of observations for each
+            band, if available in the input object file
+        nobs_tot_col: 'str', optional
+            Identifies which column contains the total number of observations,
+            if available in the input object file
+        provenance_col: 'str', optional
```
drewoldag, line 35: nit: `provenance_col` is out of order compared to the arguments in the function definition.

src/lsstseries/utils/column_mapper/column_mapper.py
```diff
@@ -0,0 +1,200 @@
...
+            Identifies which mapping scheme to use
+
+        Returns
+        -------
+        A ColumnMapper subclass object dependent on the map_id provided, for example
+        ZTFColumnMapper in the case of "ZTF"
+
+        """
+        if map_id in self.known_maps.keys():
+            if map_id == "ZTF":
```
drewoldag, line 92: I don't think you need the conditional check here. Can this be reduced to:
```
if map_id in self.known_maps:
    return self.known_maps[map_id]()._set_known_map()
else
    ...
```

Perhaps it's less clear this way ☝️ , if so, feel free to leave as is. 
Either way, it would be nice to avoid having to extend the conditional for each new `map_id`.

src/lsstseries/utils/column_mapper/column_mapper.py
```diff
@@ -0,0 +1,200 @@
...
+            Indicates whether to also return a list of missing columns
+
+        Returns
+        -------
+        `bool` or tuple of (bool, list) dependent on show_needed parameter
+
+        """
+
+        # Grab required column keys
+        required_keys = [item[0] for item in self.required.items() if item[1]]
```
drewoldag, line 113: This is a little difficult to read, and requires that the reader know that `self.required` is a dictionary of `str : bool` key/values. Certainly feels like overkill to introduce a new miniclass `Column` with a `name` and `is_required`. But I do wonder if there's a way to make this more intuitive for future devs.

One option that is much more lightweight than a whole class or data class is a named tuple. I looked over the code and didn't immediately see places where we make use of the properties of a dictionary. Perhaps you could consider in the `__init__` method:

```
from collections import namedtuple
Column = namedtuple('Column', ['name', 'is_required'])

self.required = [
            Column(name="id_col", is_required=True),
            Column("flux_col", True),  # `name` and `is_required` aren't necessary to instantiate
            ...
        ]
```

Then here, you can do this:
```
required_keys = [col.name for col in self.required if col.is_required]
```

src/lsstseries/analysis/structure_function/__init__.py
```diff
@@ -1,10 +1,11 @@
-from .base_argument_container import StructureFunctionArgumentContainer
-from .base_calculator import StructureFunctionCalculator, register_sf_subclasses
-from .basic.calculator import BasicStructureFunctionCalculator
-from .bauer_2009a.calculator import Bauer2009AStructureFunctionCalculator
-from .bauer_2009b.calculator import Bauer2009BStructureFunctionCalculator
-from .macleod_2012.calculator import Macleod2012StructureFunctionCalculator
-from .schmidt_2010.calculator import Schmidt2010StructureFunctionCalculator
+from .base_argument_container import StructureFunctionArgumentContainer  # noqa
```
drewoldag, line 1: Thank you!

drewoldag (approved):
Looks pretty cool to me. I left a couple of comments, none that are blocking though.

Issue #65, opened 2023-04-05
Basic API improvements - core data frame functions


nevencaplar:
Implement API for basic dataframe functions. These should include 

- select
- filter
- band select (select data in a given band)
- band count (number of observations in a given band)

jeremykubica:
Closed by #120 and #117.

Issue #44, opened 2023-03-02
Filtering and Column Arithmetic


dougbrn:
Filtering the ensemble on a given column is a core requirement for the ensemble actually being useful. As a somewhat related extension, the ability to add new columns given arithmetic on existing columns (or any criteria) is desirable. These two things will enable useful functionality such as color cuts, which would be implemented as a function that creates a color column and then filters on some color threshold.

dougbrn:
My thinking is that whatever we choose to do should be consistent across the functions that modify the object/source tables. Currently, that is more to the alternative, where ens.filter would update ens._source or ens._object directly, though we also return the ensemble object back as well. The advantage of it being that a user avoids the subtle issue of doing something like:
```
ens.filter()

ens.info() # Why didn't my filter work??
```

Returning a view may also be in light conflict with dask and it's lazy loading, as we would just return the lazy dataframe object view and the user would still need to call a compute() to actual execute the task graph.

jeremykubica:
How should the criteria specification work on this. I can imagine a few scenarios:
1) Predefined criteria (equals, less than, etc.)
2) Passing a function to be evaluated on each row
3) Passing an arbitrary string with the condition, allowing things like `df[df["col_name1"] < 3 or df["col_name2"] == "Hello"]`

What would be most useful here?

PR #117, opened 2023-05-26
Add filter operations

jeremykubica:
Add filter operations for the tables by wrapping Dask DataFrame's `query` function and the ability to pass a series of Booleans.

Also adds a `_lazy_sync_table` function to simplify the logic when you only need to modify one table.

### DIFF
--- a/docs/notebooks/working_with_the_ensemble.ipynb
+++ b/docs/notebooks/working_with_the_ensemble.ipynb
@@ -1,7 +1,6 @@
 {
  "cells": [
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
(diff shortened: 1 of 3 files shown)


jeremykubica:
Addresses part of #44 
Addresses part of #65

dougbrn:
Thanks @jeremykubica, this looks good. Would you mind adding a simple example to the docs as part of the PR? We have a space for it here: https://github.com/lincc-frameworks/lsstseries/blob/main/docs/notebooks/working_with_the_ensemble.ipynb?short_path=f8e5689#L224

dougbrn (approved):
Nice, looks good!

PR #118, opened 2023-05-26
Change name of the filter function to query

jeremykubica:
Change name of `filter` to `query` because it is just wrapping Dask's `query` function.


dougbrn (approved):
looks good to me, thanks @jeremykubica!

PR #119, opened 2023-05-26
Add a persist function to Ensemble

jeremykubica:
Calls Dask's persist on each of the ensemble's tables.

### DIFF
--- a/docs/notebooks/working_with_the_ensemble.ipynb
+++ b/docs/notebooks/working_with_the_ensemble.ipynb
@@ -120,7 +120,7 @@
    "source": [
     "With this compute, we see above that we now have a populated dataframe (a Pandas dataframe in fact!). From this, many workflows in Dask and by extension lsstseries, will look like a series of lazily evaluated commands that are chained together and then executed with a .compute() call at the end of the workflow.\n",
     "\n",
-    "Alternatively we can use `ens.persist()` to execute the chained commands without loading anything into memory. This can speed up future `compute()` calls."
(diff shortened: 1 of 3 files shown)


docs/notebooks/working_with_the_ensemble.ipynb
```diff
@@ -118,7 +118,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "With this compute, we see above that we now have a populated dataframe (a Pandas dataframe in fact!). From this, many workflows in Dask and by extension lsstseries, will look like a series of lazily evaluated commands that are chained together and then executed with a .compute() call at the end of the workflow."
+    "With this compute, we see above that we now have a populated dataframe (a Pandas dataframe in fact!). From this, many workflows in Dask and by extension lsstseries, will look like a series of lazily evaluated commands that are chained together and then executed with a .compute() call at the end of the workflow.\n",
+    "\n",
+    "Alternatively we can use `ens.persist()` to execute the chained commands without loading anything into memory. This can speed up future `compute()` calls."
```
dougbrn, line 123: nit: a user might take "without loading anything into memory" to imply that nothing is ever in memory, but persist does work on the data in memory as the task graph is executing. Maybe "without loading the result into memory" would be clearer.

dougbrn (approved):
One nit, but looks good overall

docs/notebooks/working_with_the_ensemble.ipynb
jeremykubica: That makes sense. I reworded it using your suggestion.

PR #120, opened 2023-05-30
Add a basic select function

jeremykubica:
Add a select function that takes a list of columns to keep.


dougbrn (approved):
Looks good to me!

PR #123, opened 2023-05-30
Building out SF documentation

drewoldag:
This PR extends the existing Structure Function tutorial notebook. 

In the process of extending the tutorial, I identified a few rough edges that are also fixed in this PR. Most notably is moving the `SF_METHODS` global variable out of the original __init__.py file, and into it's own module, `calculator_registrar.py`. Along with the global variable, I moved one of the supporting methods out of `base_calculator.py` into `calculator_registrar.py`. 

While doing that I also moved a couple of tests to maintain symmetry, and added a test for a little more coverage.


src/lsstseries/analysis/structure_function/base_calculator.py
drewoldag, line 1: Moved this function to `calculator_registrar.py`.

docs/notebooks/working_with_structure_function.ipynb
```diff
@@ -32,12 +32,23 @@
    "cell_type": "markdown",
```
dougbrn, line 32: <p>grammar nit: "methods will are"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/lsstseries/pull/123/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

dougbrn:
I think what you've added docs wise is fantastic, thanks a ton for this! _At some point_, we should probably add even a bit more to the structure function docs providing description of what the function is doing scientifically. We have a lot of docs right now on how to interact with the API, but at the end of the day we will really want the achievable science to be front and center in our docs (@nevencaplar feel free to chime in on this). But again, that's very much future work and this looks great. Let me know when you take this out of draft mode and I'll take another look/approve!

docs/notebooks/working_with_structure_function.ipynb
drewoldag: Updated!

dougbrn (approved):
Looks great!

PR #124, opened 2023-05-31
Update minimum Dask version

jeremykubica:
One of the tests started crashing when npartitions=2 and repeated ids were used. This appears to be due to a known bug in Dask (https://github.com/dask/dask/issues/10118) that has since been corrected. This PR updates the minimum version of Dask to pick up the fix.


smcguire-cmu (approved):
Looks good to me

Issue #64, opened 2023-04-05
Basic API improvements - generate new columns


nevencaplar:
Given the results of some analysis, have an API or ``easy'' ability to append the new results to the ensamble table (object table) as a custom column.

drewoldag:
Just read this and I don't have any context, but the approach that works well in relational databases is to have a couple of thin ancillary tables indexed on `Object.id`. 
```
**ancillary_data**
object_id BIGINT,
data_type_id INT,
value <type>

**data_types**
data_type_id INT,
data_type_representation <what type of data, INT, FLOAT, CHAR, ...>,
description <STRING>
```

dougbrn:
Assign addresses this, we may have more structured functions as well but I think the spirit of this issue is addressed. Feel free to reopen if you disagree :)

PR #125, opened 2023-05-31
Wrap Dask's Assign function

### LINKED ISSUE #44: Filtering and Column Arithmetic

jeremykubica:
Wrap Dask's `assign` function to allow users to construct new columns in either the source or object table.

Closes #44 
Also addresses part of #64

### DIFF
--- a/docs/notebooks/working_with_the_ensemble.ipynb
+++ b/docs/notebooks/working_with_the_ensemble.ipynb
@@ -311,6 +311,27 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "## Assignments and Column Manipulation\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


dougbrn (approved):
Nice, this looks great! Approved.

Aside: I think this motivates an Ensemble.update_column_map function, where you can update the internal mapping of critical columns. For example, a user makes this new column, and now they want it to be treated as the flux column from now on. I'll make an issue for this

PR #127, opened 2023-05-31
Create `LightCurve` base class for analysis functions

drewoldag:
Created `LightCurve` base class that `StructureFunctionLightCurve` inherits from.
Updated references and tests.

### DIFF
--- a/src/lsstseries/analysis/light_curve.py
+++ b/src/lsstseries/analysis/light_curve.py
@@ -1,15 +1,38 @@
-from typing import Optional
 import numpy as np
-
-MIN_OBSERVATIONS_REQUIRED_FOR_SF = 3
 
 
 class LightCurve:
-    def __init__(self, times: np.ndarray, fluxes: np.ndarray, errors: np.ndarray):
+    """This base class is meant to support various analysis routines and be
+    extended as needed. (Hence it's location in the `analysis` package.)
+
+    The base class ensures that the data for a single lightcurve is well formed.
+    Namely that the input data is all of the same length, with NaN's removed and
+    that there are enough observations to perform a given analysis.
+    """
+
+    def __init__(
+        self, times: np.ndarray, fluxes: np.ndarray, errors: np.ndarray, minimum_observations: int = 0
+    ):
+        """Calls the processing functions to clean and validate the input data.
+        All of the input arrays should have the same length.
+
+        Parameters
+        ----------
+        times : np.ndarray
+            The timestamps values of the data.
+        fluxes : np.ndarray
+            The flux/magnitude values of the data.
+        errors : np.ndarray
+            The error in the measurements.
+        minimum_observations : int, optional
+            The minimum number of data points needed to successfully process
+            the lightcurve, by default 0
+        """
         # The primary data
         self._times = times
         self._fluxes = fluxes
         self._errors = errors
+        self._minimum_observations = minimum_observations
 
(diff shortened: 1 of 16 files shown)


src/lsstseries/analysis/light_curve.py
```diff
@@ -0,0 +1,49 @@
...
+        self._times = self._times[~nan_mask]
+        self._fluxes = self._fluxes[~nan_mask]
+        self._errors = self._errors[~nan_mask]
+
+    def _check_input_data_size_is_equal(self):
+        """Make sure that the three input np.arrays have the same size"""
+        if self._times.size != self._fluxes.size or self._times.size != self._errors.size:
+            raise ValueError("Input np.arrays are expected to have the same size.")
+
+    def _check_input_data_length_is_sufficient(self):
```
jeremykubica, line 44: If the goal is to have this support more than just the structure function, then I would suggest making the minimum number of points an optional parameter. You could set the default to `MIN_OBSERVATIONS_REQUIRED_FOR_SF`.

src/lsstseries/analysis/light_curve.py
```diff
@@ -0,0 +1,49 @@
+from typing import Optional
+import numpy as np
+
+MIN_OBSERVATIONS_REQUIRED_FOR_SF = 3
```
jeremykubica, line 4: Can this be moved into the subclass and passed into `LightCurve`?

src/lsstseries/analysis/light_curve.py
```diff
@@ -0,0 +1,49 @@
+from typing import Optional
```
jeremykubica, line 1: Comment on the high level placement of this file (not this specific line): Would this make sense to go in top-level src directory as one of the fundamental data types?

src/lsstseries/analysis/light_curve.py
drewoldag: Yep, good call. I'll move this out of the base class and into the `StructureFunctionLightCurve` class.

src/lsstseries/analysis/light_curve.py
drewoldag: I can add a docstring for this class to explain the motivation for putting it here. The goal is for this to be a container class for use in analysis functions. So I _think_ that keeping it here makes sense. But if there are strong opinions, I'm happy to adjust.

jeremykubica: I don't feel strongly about this. So feel free to keep it where it is

src/lsstseries/analysis/light_curve.py
drewoldag: This makes sense to me. I've added an input argument for the base class with a default of 0. The subclass will provide `MIN_OBSERVATIONS_REQUIRED_FOR_SF` (equal to 3) when calling `super().__init__(...)`. 

Thanks for the suggestion!

src/lsstseries/analysis/light_curve.py
drewoldag: Gotcha, for now, kept where it is.

src/lsstseries/analysis/light_curve.py
drewoldag: I've moved this constant to the `StructureFunctionLightCurve` class, and pass it as a parameter back to the base class which will do the array length check. Thanks for the suggestion here.

PR #129, opened 2023-06-02
Add the ability to change column mappings in an Ensemble

### LINKED ISSUE #126: Updating the Column Map
With the addition of Ensemble.assign, users can now add columns derived from other columns. It's reasonable to expect a user to assign a new column, then want to replace an existing critical column with their new column. For example, a modified flux or shifted time column.

We should add an Ensemble.update_column_map function, which takes in a new ColumnMapper object or just the direct kwargs

jeremykubica:
Closes #126

### DIFF
--- a/src/lsstseries/ensemble.py
+++ b/src/lsstseries/ensemble.py
@@ -718,8 +718,68 @@ class Ensemble:
             **kwargs,
         )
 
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
looks great, thanks!

PR #133, opened 2023-06-07
update project name for PyPI

dougbrn:
"tape" is taken, "lincc-tape" is my proposed alternative


docs/conf.py
```diff
@@ -16,10 +16,10 @@
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
-project = "tape"
+project = "lincc-tape"
```
jeremykubica, line 19: Let's go with `lf-tape` instead

jeremykubica (changes requested):

jeremykubica (approved):

PR #134, opened 2023-06-07
Create stub notebook for Structure Function science showcase.

drewoldag:
This is really meant to be a stub notebook. The goal is to make sure EzTao can be imported and used to create some example data. Currently, EzTao _doesn't_ work, but there's a fix that should be made shortly. I've copied some data from a manual run of EzTao into the stub to show how to run calc_sf2 with various SF calculation methods and plotted those for comparison against 1 light curve. 

Also there's some stub cells in there for working with an Ensemble. Again, when EzTao is updated, we'll be able to make this section much cleaner.

Again, this stub is here as a starting point. It's by not meant to cover all the examples called out in this doc: https://docs.google.com/document/d/1pKxhAZHyLOkXED2EqmYP1nQHZe_FJLfgGTjsnVY80Oo/edit#heading=h.3e7qxfyqkdby


docs/notebooks/structure_function_showcase.ipynb
```diff
@@ -0,0 +1,680 @@
+{
```
dougbrn, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    import matplotlib.pyplot as pylt
</pre><p>My understanding is that generally "plt" is the preferred shorthand here, not "pylt"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/134/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

dougbrn, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.        id = np.array([[1],[2]]),
</pre><p>id is a native python function, we should try to avoid overwriting the namespace</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/134/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

dougbrn, line 1: <pre class="ql-syntax" spellcheck="false">Line #6.    # ens.from_source_dict(source_dict=data_dict, column_mapper=manual_colmap)
</pre><p>As we discussed directly, this fails because 2d arrays are attempting to be passed along</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/134/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

drewoldag: Updated!

drewoldag: Fixed this.

dougbrn:
All of the demo contents here look good. I do have to ask though before approving, do you think it makes sense to hold off on merging this in until eztao is properly updated? Maybe that was already the plan and I'm just out of the loop.

drewoldag:
I'm not a strong advocate either way. Happy to leave this unmerged until then if that make sense.

Additionally if someone else is going to pick up the work to fill this out with all the requests that Neven has called out, then I would be fine with them just taking over this PR.

dougbrn:
Okay, I _think_ I'm leaning towards holding off on merging this until eztao is functional for this notebooks purpose. Not sure if @nevencaplar has an opinion here, mine definitely isn't very strong. 😄

docs/notebooks/structure_function_showcase.ipynb
drewoldag: Changed this to `lc_id`.

dougbrn (approved):
see that this has a working eztao now, nice looks good.

PR #135, opened 2023-06-12
Remove per-analysis logic from Ensemble

jeremykubica:
This PR is to start discussion. I would like to remove as much of the per-analysis logic as possible from the `Ensemble` class. This represents one possible way to do this, but I'm not sure it is the best.

Create helper functions to provide the required columns (args) and meta for each analysis. Use `ColumnMapper` to link these back to the correct columns in the `Ensemble`. This requires that the user does a little more set up to run the analysis.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -11,7 +11,7 @@ import autoapi
 from importlib.metadata import version
 
 # Define path to the code to be documented **relative to where conf.py (this file) is kept**
-sys.path.insert(0, os.path.abspath('../src/'))
+sys.path.insert(0, os.path.abspath("../src/"))
 
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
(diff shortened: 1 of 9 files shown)


dougbrn:
We talked about this a bit, but to summarize here; I think there's a lot of value in what's being proposed from a maintainability standpoint. I'm not keen on the extra functions (such as meta_sf2, etc.) that are required to be imported and supplied. I think that users would benefit a lot from the Ensemble continuing to handle these things under the hood. So maybe an alternative approach is good? I know you had a few ideas, the one I proposed was to have the meta and required columns live in the analysis function files (as presented here), but continue to map them within ensemble.batch. This would alleviate some of the concern with the Ensemble object becoming too complex, as at least in this instance we modify the mappings within the analysis function file. But I'm happy with other approaches that keep the user feel at a similar level to how things are now.

jeremykubica:
Reverting pull request since this isn't the approach we are going to go with. Thanks for taking a look and the discussion.

PR #138, opened 2023-06-23
Read datasets

dougbrn:
Implements two functions that support the tape_benchmarking datasets. The first retrieves the set of available datasets to read and the second provides the ability to load them into an ensemble object.

### DIFF
--- a/tests/tape_tests/test_ensemble.py
+++ b/tests/tape_tests/test_ensemble.py
@@ -67,6 +67,64 @@ def test_from_parquet(data_fixture, request):
         assert parquet_ensemble._source[col] is not None
 
 
+def test_available_datasets(dask_client):
+    """
+    Test that the ensemble is able to successfully read in the list of available TAPE datasets
+    """
+    ens = Ensemble(client=dask_client)
+
+    datasets = ens.available_datasets()
+
+    assert isinstance(datasets, dict)
+    assert len(datasets) > 0  # Find at least one
+
+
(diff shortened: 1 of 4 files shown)


tests/tape_tests/test_ensemble.py
```diff
@@ -67,6 +67,64 @@ def test_from_parquet(data_fixture, request):
         assert parquet_ensemble._source[col] is not None
 
 
+def test_available_datasets(dask_client):
+    """
+    Test that the ensemble is able to successfully read in the list of available TAPE datasets
+    """
+    ens = Ensemble(client=dask_client)
+
+    datasets = ens.available_datasets()
```
jeremykubica, line 76: Best practices is usually to avoid external dependencies in unit tests because they increase run time and have the potential for flaky tests (if the external dependency is down). The alternative here would be to mock out the data set information and get call, which adds a lot of complexity. So I think we can leave this in there for now. But we should think if there is an alternative approach we can use for testing in the future.

pyproject.toml
```diff
@@ -23,6 +23,8 @@ dependencies = [
     'pyvo',
     'scipy',
     'matplotlib',
+    'requests',
+    'aiohttp',
```
jeremykubica, line 27: Is `aiohttp` needed? I do not see it used anywhere.

jeremykubica (approved):
One structural comment to think about for future changes.

PR #155, opened 2023-07-11
Issue74 improve dropna

wenneman:
Added the ability to drop NA values from either the source or object table, and now leveraging the Dask Data Frame dropna() function passing along whatever arguments are provided.

### DIFF
--- a/docs/notebooks/working_with_the_ensemble.ipynb
+++ b/docs/notebooks/working_with_the_ensemble.ipynb
@@ -267,7 +267,8 @@
    "outputs": [],
    "source": [
     "# Cleaning nans\n",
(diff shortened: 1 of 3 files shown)


dougbrn:
@wenneman, this looks good! Two minor things, the first is that the docs build is failing for this PR. This is because there's a dropna call using the old API in this notebook: https://github.com/lincc-frameworks/tape/blob/main/docs/notebooks/working_with_the_ensemble.ipynb. This should be changed to use the new Dask arguments, probably using how="any". The other thing is the slight coverage loss in the CI, I think this is just because you've added an exception when "source" or "object" is not specified as an input, maybe just add something that exercises the fail case?

dougbrn (approved):
looks good, feel free to merge

PR #161, opened 2023-07-14
Workflow to submit benchmarks to Sasquatch.

camposandro:
Creates a GitHub workflow that:
- Computes runtime benchmarks for a project using [pytest-benchmark](https://pytest-benchmark.readthedocs.io/en/latest/)
- Invokes a custom [Python script](https://github.com/lincc-frameworks/benchmarking-scripts/blob/metrics-script/metrics.py) to extract relevant metrics and submit them to Sasquatch

**TODO**
- [ ] Update workflow trigger events (pushes to master, smoke test, on release)
- [ ] Investigate if / how to send multiple metrics for a module at once
- [ ] Investigate how to compute memory usage


dougbrn:
Closing as we'll eventually get benchmarks via the PPT. @camposandro if there's a reason why this should be open feel free to re-open!

PR #167, opened 2023-07-18
downgrade numpy for numba support

dougbrn:
readthedocs has been failing builds since the sf showcase notebook went in (slightly concerning that our CI didn't catch this), at least one layer of this is that Numba needs Numpy <= 1.24.0, updated that for the docs requirements.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,5 +6,5 @@ ipython
 jupytext
 jupyter
 matplotlib
-numpy==1.24.0
+numpy==1.23.5
 eztao
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -6,5 +6,5 @@ ipython
 jupytext
 jupyter
 matplotlib
-numpy
+numpy==1.24.0
 eztao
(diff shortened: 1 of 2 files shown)


drewoldag (approved):
This looks reasonable. Since the CI wasn't really helping before, was there a change that we should make to help out in the future?

dougbrn:
> This looks reasonable. Since the CI wasn't really helping before, was there a change that we should make to help out in the future?

So I think the difference between CI and RTD is that CI installs the main TAPE dependencies first, then the doc dependencies from docs/requirements.txt. While it's probably ideal to have these two doing the same thing, I think the potentially fine solution here is just to fix the main numpy version (in pyproject.toml) to 1.23.5 and remove the doc requirement. I've just done that in the most recent commit. I was originally hesitant to limit the main doc version based on this numba incompatibility in the docs, but it's probably better that all our various code is set to use the same dependencies. If numba/eztao cause a lot of lag in dependency compatibility going forward it may also be worth thinking about if we actually want these supported in our codebase/docs.

PR #168, opened 2023-07-18
SF^1 to SF^2 bug fix

nevencaplar:
Fixes https://github.com/lincc-frameworks/tape/issues/165, makes all calculators to return SF^2, instead of SF^1 for ``non-basic'' calculators.


drewoldag (approved):
Looks good. Sorry about introducing these bugs!

dougbrn (changes requested):
Going off the failed CI, it looks like some of the unit tests will need to be updated, as they are testing for specific sf output values that have been changed now that the functions are returning the previous result squared

dougbrn (approved):
See that this is all good on the CI front now, looks good.

Issue #176, opened 2023-07-20
Ensemble.coalesce


dougbrn:
Motivated by #172, we should make a coalesce function that takes several columns that represent the same quantity, but for different groupings of rows, and returns a single new column with quantities for all rows. The most practical example is multiple input surveys, where we might have ztf_flux and ps1_flux as columns in the same source table, since the provenance of a given row will always only have a non-nan value for just one of these columns, the combination should be doable via a coalescing function.

dougbrn:
This is very likely needed for #140

dougbrn:
Closed by #178

PR #178, opened 2023-07-21
add coalesce

dougbrn:
Adds the coalesce function, as described in #176

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -410,6 +410,83 @@ class Ensemble:
             raise ValueError(f"{table} is not one of 'object' or 'source'")
         return self
 
+    def coalesce(self, input_cols, output_col, table="object", drop_inputs=False):
+        """Combines multiple input columns into a single output column, with
+        values equal to the first non-nan value encountered in the input cols.
+
+        Parameters
+        ----------
+        input_cols: `list`
+            The list of column names to coalesce into a single column.
+        output_col: `str`, optional
+            The name of the coalesced output column.
+        table: `str`, optional
+            "source" or "object", the table in which the input columns are
+            located.
+        drop_inputs: `bool`, optional
+            Determines whether the input columns are dropped or preserved. If
+            a mapped column is an input and dropped, the output column is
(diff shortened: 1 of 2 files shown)


dougbrn:
Unit tests are failing in line with fails from the smoke tests last night, something is weird about the calc_stetson_J dictionaries. I will investigate and include whatever change is needed for the smoke tests as part of this PR.

dougbrn:
Changed my mind on how to approach the smoke tests, it's going in separately in PR #179 instead.

src/tape/ensemble.py
```diff
@@ -410,6 +410,83 @@ def assign(self, table="object", **kwargs):
...
+        table_ddf = table_ddf.assign(**{output_col: coalesce_col})
+
+        # Drop the input columns if wanted
+        if drop_inputs:
+            # First check to see if any dropped columns were critical columns
+            current_map = self.make_column_map().map
+            cols_to_update = [key for key in current_map if current_map[key] in input_cols]
+
+            # Update critical columns to the new output column as needed
+            if len(cols_to_update) != 0:
```
drewoldag, line 462: I think that this can be just `if len(cols_to_updates):`

src/tape/ensemble.py
```diff
@@ -410,6 +410,83 @@ def assign(self, table="object", **kwargs):
...
+
+                new_colmap = self.make_column_map()
+                new_colmap.map = new_map
+
+                # Update the mapping
+                self.update_column_mapping(new_colmap)
+
+            # Theoretically a user could assign multiple critical columns in the input cols, this is very
+            # likely to be a mistake, so we throw a warning here to alert them.
+            if len(cols_to_update) > 1:
```
drewoldag, line 475: Would it be worth moving this conditional to just after line 459? Perhaps it's  not important since this is just a warning.

drewoldag (approved):
This looks pretty solid to me. I left a couple small comments, but otherwise it seems good to go.

dougbrn:
Thanks for the suggestions, implemented both!

PR #188, opened 2023-07-29
Add support for Python 3.11 on Github Actions

### LINKED ISSUE #187: Update github actions to test python 3.11
With Python 3.11 being fully supported since 2022-10-24, we should introduced testing against 3.11 version in our current testing suite.

aitorres:
Closes #187

## What I did?

Added Python 3.11 to the version matrix on Github Actions' config files.

## How to test?

All Github Actions checks should succeed on this PR, including new checks for linting, testing / coverage and smoke tests on Python 3.11

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
 
     steps:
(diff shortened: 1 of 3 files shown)


nevencaplar:
@dougbrn  I do think this is as simple as adding 3.11 in github actions. Just checking that we are not missing something?

nevencaplar (approved):
@dougbrn I do think this is as simple as adding 3.11 in github actions. Just checking that we are not missing something?

dougbrn (approved):
Yeah I think that would be it @nevencaplar. Thanks @aitorres !

PR #189, opened 2023-08-02
Dedicated error estimate in structure function

### LINKED ISSUE #180: Error estimate in structure function should not require equally_weight_lightcurves=True
At the moment, in order to bootstrap structure function calculation one needs to specify `equally_weight_lightcurves=True`.  For example see this [test permalink](https://github.com/lincc-frameworks/tape/blob/f96366996b53fb465c45ea350217097636bebd82/tests/tape_tests/test_analysis.py#L1102C5-L1102C5) and this code that invokes [select_difference_sample](https://github.com/lincc-frameworks/tape/blob/f96366996b53fb465c45ea350217097636bebd82/src/tape/analysis/structure_function/base_calculator.py#L52) function that does actually random sampling. This is also in the case that an user wants to analyze only a single lightcurve. 

This argument should only be needed when `combine=True`, and only if an user want to actually treat all light-curves equally, which may not be the case.

nevencaplar:
Fixes #180 

## What did I do?

- Introduced `estimate_err` argument; when it is true the errors are estimated via bootstraping
    - do 100 repetitions by default  
- Renamed `_equally_weight_lightcurves` function to `_bootstrap` function

- Changed behavior depending on the bool value of `_equally_weight_lightcurves` argument
     - if self `_equally_weight_lightcurves = True` then go and take `[min(self._difference_values_per_lightcurve)]` on from each lightcurve
    - if self `_equally_weight_lightcurves = False` then select all of the differences avaliable
- Modified test results

### DIFF
--- a/src/tape/analysis/structure_function/base_argument_container.py
+++ b/src/tape/analysis/structure_function/base_argument_container.py
@@ -68,7 +68,7 @@ class StructureFunctionArgumentContainer:
         calculation_repetitions: `int`, optional
             Specifies the number of times to repeat the structure function
             calculation. Typically this would be used when setting
-            `estimate_err = True`. By default 1 when not estimating erros,
+            `estimate_err = True`. By default 1 when not estimating errors,
             and 100 when estimating errors.
         lower_error_quantile: `float`, optional
             When calculation_repetitions > 1 we will calculate the
--- a/src/tape/analysis/structure_function/base_argument_container.py
+++ b/src/tape/analysis/structure_function/base_argument_container.py
@@ -61,10 +61,15 @@ class StructureFunctionArgumentContainer:
             If a value is not provided here, then the default number of
             lightcurve samples will be equal to the least number of differences
             in the available lightcurves. By default None.
+        estimate_err: `bool`, optional
+            Specifies if the structure function errors are to be estimated,
+            via bootstraping the sample and taking note of 16 and 84 percentile
+            of the resulting distribution
         calculation_repetitions: `int`, optional
             Specifies the number of times to repeat the structure function
             calculation. Typically this would be used when setting
-            `equally_weight_lightcurves = True`. By default 1.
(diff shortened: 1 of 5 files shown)


nevencaplar:
Closes #180

## What did I do?

- Introduced `estimate_err` argument; when it is true the errors are estimated via bootstraping
    - do 100 repetitions by default  
- Renamed `_equally_weight_lightcurves` function to `_bootstrap` function

- Changed behavior depending `_equally_weight_lightcurves` argument
     - if self `_equally_weight_lightcurves = True` then go and take [min(self._difference_values_per_lightcurve)] on (https://github.com/lincc-frameworks/tape/blob/f96366996b53fb465c45ea350217097636bebd82/src/tape/analysis/structure_function/base_calculator.py#L47C87-L47C87) from each lightcurve
    - if self `_equally_weight_lightcurves = False` then select all of the differences avaliable
- Modified test results

jeremykubica (approved):

src/tape/analysis/structure_function/base_argument_container.py
```diff
@@ -61,10 +61,15 @@ class StructureFunctionArgumentContainer:
...
             in the available lightcurves. By default None.
+        estimate_err: `bool`, optional
+            Specifies if the structure function errors are to be estimated,
+            via bootstraping the sample and taking note of 16 and 84 percentile
+            of the resulting distribution
         calculation_repetitions: `int`, optional
             Specifies the number of times to repeat the structure function
             calculation. Typically this would be used when setting
-            `equally_weight_lightcurves = True`. By default 1.
+            `estimate_err = True`. By default 1 when not estimating erros,
```
dougbrn, line 71: nit: erros->errors

dougbrn (approved):
Looks good to me!

drewoldag (approved):
This looks good to me. Thanks for cleaning this up!

Issue #183, opened 2023-07-27
Add a Flux to Magnitude Conversion Function


dougbrn:
Mag = -2.5*np.log10(flux) + zp if zp is an integer of order ~10, it can also be Mag = -2.5*np.log10(flux/zp) if zp is a small fraction.

dougbrn:
This is a more complex problem than just implementing the above equations. The astropy docs show their ecosystem for doing unit conversions: https://docs.astropy.org/en/stable/units/logarithmic_units.html

We could try to wrap this, but it may not end up being that helpful to users. Instead, we should maybe consider just making sure that users are able to work with a column and assign new columns based on the result. In other words, let users go to Astropy to do what Astropy does best rather than try to stand something up on top of Astropy.

dougbrn:
We also need to make sure that errors are propagated.

PR #190, opened 2023-08-07
Flux to Mag Converter Function

dougbrn:
Adds a function for converting a flux column to a magnitude column. Requires a zero point column to be additionally supplied. Errors are propagated if a error column is provided. Implements #183.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1095,7 +1095,7 @@ class Ensemble:
         return self
 
     def convert_flux_to_mag(self, flux_col, zero_point, err_col=None, zp_form="mag", out_col_name=None):
-        """Converts a flux column into a magnitude.
+        """Converts a flux column into a magnitude column.
 
         Parameters
         ----------
@@ -1106,12 +1106,15 @@ class Ensemble:
             information for column transformation.
         err_col: 'str', optional
             The name of the ensemble column containing the errors to propagate.
-        zp_form: `str`
+            Errors are propagated using the following approximation:
+            Err= (2.5/log(10))*(flux_error/flux), which holds mainly when the
+            error in flux is much smaller than the flux.
+        zp_form: `str`, optional
             The form of the zero point column, either "flux" or
             "magnitude"/"mag". Determines how the zero point (zp) is applied in
             the conversion. If "flux", then the function is applied as
-            Mag=-2.5*log10(flux/zp), or if "magnitude", then
-            Mag=-2.5*log10(flux)+zp.
+            mag=-2.5*log10(flux/zp), or if "magnitude", then
+            mag=-2.5*log10(flux)+zp.
         out_col_name: 'str', optional
             The name of the output magnitude column, if None then the output
             is just the flux column name + "_mag". The error column is also
@@ -1120,7 +1123,7 @@ class Ensemble:
         Returns
         ----------
         ensemble: `tape.ensemble.Ensemble`
-            The ensemble object with a new mag column.
+            The ensemble object with a new magnitude (and error) column.
 
         """
         if out_col_name is None:
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1094,16 +1094,18 @@ class Ensemble:
         self._object_dirty = False
         return self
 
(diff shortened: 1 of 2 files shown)


src/tape/ensemble.py
```diff
@@ -1094,6 +1094,59 @@ def from_source_dict(self, source_dict, column_mapper=None, npartitions=1, **kwa
...
+        Parameters
+        ----------
+        flux_col: 'str'
+            The name of the ensemble flux column to convert into magnitudes.
+        zero_point: 'str'
+            The name of the ensemble column containing the zero point
+            information for column transformation.
+        err_col: 'str', optional
+            The name of the ensemble column containing the errors to propagate.
+        zp_form: `str`
```
nevencaplar, line 1109: include `,optional`

src/tape/ensemble.py
```diff
@@ -1094,6 +1094,59 @@ def from_source_dict(self, source_dict, column_mapper=None, npartitions=1, **kwa
...
+                **{out_col_name: lambda x: -2.5 * np.log10(x[flux_col]) + x[zero_point]}
+            )
+
+        else:
+            raise ValueError(f"{zp_form} is not a valid zero_point format.")
+
+        # Calculate Errors
+        if err_col is not None:
+            self._source = self._source.assign(
+                **{out_col_name + "_err": lambda x: (2.5 / np.log(10)) * (x[err_col] / x[flux_col])}
```
nevencaplar, line 1145: I would add a note or a comment in the docstring to note that this is only correct if the limit where the error on the flux is much smaller than the flux itself. This is not an exact formula, and that should be noted somewhere.

src/tape/ensemble.py
```diff
@@ -1094,6 +1094,59 @@ def from_source_dict(self, source_dict, column_mapper=None, npartitions=1, **kwa
         self._object_dirty = False
         return self
 
+    def convert_flux_to_mag(self, flux_col, zero_point, err_col=None, zp_form="mag", out_col_name=None):
+        """Converts a flux column into a magnitude.
```
nevencaplar, line 1098: ...into a magnitude column

src/tape/ensemble.py
```diff
@@ -1094,6 +1094,59 @@ def from_source_dict(self, source_dict, column_mapper=None, npartitions=1, **kwa
...
+        Returns
+        ----------
+        ensemble: `tape.ensemble.Ensemble`
+            The ensemble object with a new mag column.
+
+        """
+        if out_col_name is None:
+            out_col_name = flux_col + "_mag"
+
+        if zp_form == "flux":  # mag = -2.5*np.log10(flux/zp)
```
nevencaplar, line 1129: mag -> Mag to be consistent with the docstring (or change in docstring); just to be consistent

src/tape/ensemble.py
```diff
@@ -1094,6 +1094,59 @@ def from_source_dict(self, source_dict, column_mapper=None, npartitions=1, **kwa
...
+            self._source = self._source.assign(
+                **{out_col_name: lambda x: -2.5 * np.log10(x[flux_col] / x[zero_point])}
+            )
+
+        elif zp_form == "magnitude" or zp_form == "mag":  # mag = -2.5*np.log10(flux) + zp
```
nevencaplar, line 1134: same comment as mag <-> Mag as above

src/tape/ensemble.py
dougbrn: Will do, is there a preferable way to calculate errors here?

src/tape/ensemble.py
dougbrn: I changed the "Mag" in the equation provided in the docstring instead.

src/tape/ensemble.py
nevencaplar: I _think_ it is fine. Symmetric errors in flux create nonsymettric errors in log, i.e., one would have to create different upper and lower errors. For most of LSST like purposes I expect errors to be much smaller than the flux and the equation is ok I do not think we want to go beyond this implementation at this point.

nevencaplar (approved):

PR #197, opened 2023-08-14
Minor fixes of typos and plots for TAPE tutorial

wilsonbb:
Fixes for various typos in the TAPE tutorial notebooks.

We also try to make the plots in structure_function_showcase.ipynb more legible, and we add an additional plot to binning_slowly_changing_sources.ipynb to visualize the binning created by the calculated offset.

The bokeh package is also added to the [dev] dependencies since it's a requirement for the dask client visualizer.

### DIFF
--- a/docs/tutorials/binning_slowly_changing_sources.ipynb
+++ b/docs/tutorials/binning_slowly_changing_sources.ipynb
@@ -13,9 +13,7 @@
    "metadata": {},
    "source": [
     "In this notebook we build upon the concepts introduced in the [\"Working with the tape Ensemble object\"](https://tape.readthedocs.io/en/latest/tutorials/working_with_the_ensemble.html)\n",
-    " notebook to focus on internal helper functions for binning light curves.\n",
-    "\n",
-    "We reuse the example set of lightcurves from the previous notebook:"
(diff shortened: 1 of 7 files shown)


docs/tutorials/binning_slowly_changing_sources.ipynb
```diff
@@ -12,7 +12,8 @@
    "cell_type": "markdown",
```
dougbrn, line 12: <p>Found another needed change; we're no longer using the same set of lightcurves from the previous notebook, so the second sentence should be removed or replaced accordingly.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/197/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

dougbrn (approved):
I just found one additional needed change, but otherwise this all looks great, thanks! Feel free to merge when ready.

docs/tutorials/binning_slowly_changing_sources.ipynb
wilsonbb: Ah good catch! Opted to remove.

PR #201, opened 2023-08-22
Support of light-curve package

hombit:


pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     'coverage',
     "deprecated",
     "ipykernel", # Support for Jupyter notebooks
+    "light-curve>=0.7.3,<0.8.0",
```
dougbrn, line 30: Out of curiosity, what's the pin to <0.8.0 motivated by, is this something that we would need to change to enable more recent versions of light-curve?

tests/tape_tests/test_feature_extraction.py
```diff
@@ -0,0 +1,58 @@
...
+        "id": np.full(2 * n, 2),
+        "time": np.arange(2 * n, dtype=np.float64),
+        "flux": np.r_[np.linspace(1.0, 2.0, n), np.linspace(1.0, 2.0, n)],
+        "err": np.full(2 * n, 0.01),
+        "band": np.r_[np.full(n, "g"), np.full(n, "r")],
+    }
+    rows = {column: np.concatenate([object1[column], object2[column]]) for column in object1}
+
+    cmap = ColumnMapper(id_col="id", time_col="time", flux_col="flux", err_col="err", band_col="band")
+    ens = Ensemble().from_source_dict(rows, cmap)
```
dougbrn, line 48: In the test suite, try to make sure that any spun up ensemble objects use the dask_client defined in conftest.py, this avoids multiple clients being spun up during the test suite run

dougbrn (changes requested):
This looks good! I think the general implementation is solid, don't really have any critiques or suggestions, I did have one comment on the test suite, which I'd like to see fixed before approval. Generally, I'd also recommend adding something to the documentation on this, but were you thinking that TAPE eclipsing binary showcase would serve as a demo for using this analysis function? If that's the case I'm happy to let this move through without a tutorial for now.

pyproject.toml
hombit: I'm trying to follow [semantic versioning](https://semver.org) conventions, so potentially I can brake backward compatibility in the 0.(x+1).0 version. Practically I don't brake it in a terrible or unsound way, but I still would like the version to be restricted.

hombit:
Thanks @dougbrn! I definitely should expand docstring and add a (small) tutorial

dougbrn (approved):
Just added a fix for the strange coalesce fail, reviewed the updated PR and this all looks good!

PR #206, opened 2023-08-24
Abstract base class for analysis functions

### LINKED ISSUE #200: Base class for TAPE built-in analysis functions
Currently, `Ensemble.batch` recognizes TAPE functions by `.__name__` attribute:
https://github.com/lincc-frameworks/tape/blob/0a2957831fb83ebd7fab27b99da812ead10674e5/src/tape/ensemble.py#L721-L722

This approach has some disadvantages:
1. After a while we would have more built-in functions which would occupy the namespace and prevent user from implementing a function of the same name
2. It requires the usage of local variables `known_cols` and and `known_meta` containing some knowledge about the functions implemented in a very different part of the code, so it could cause maintenance issues 
3. This check fails if input callable has no `__name__` attribute

The last one is minor, but it is how I have actually found it:

```python
import light_curve as licu
import tape

ens = tape.Ensemble().from_dataset('s82_qso')
ens.batch(licu.Duration(), ens._time_col, ens._flux_col, ens._err_col)
```

```
AttributeError: 'light_curve.light_curve_ext.Duration' object has no attribute '__name__'
```

I believe that introduction of an abstract base class for TAPE analysis functions would solve these issues and give an optional interface for user functions:
1. We can easily check if input is a TAPE function with `isinstance`
2. We can handle `known_cols` and `known_meta` inside function 
3. User could inherit this class to create user-defined functions, or just use the interface already exists with `Ensemble.batch(func, *args)` with no worry about function name
4. This would allow creation of parametric functions

hombit:
Here I introduce an abstract base class `AnalysisFunction` for built-in analysis functions, which could also be used to create user's analysis functions. The instances of its sub-classes can be evaluated via `calculate()` method.

It also renames currently available analysis function removing `calc_` prefix from them, `calc_stetson_J -> stetson_J`, `calc_sf2 -> sf2`.

Closes #200

### DIFF
--- a/src/tape/analysis/__init__.py
+++ b/src/tape/analysis/__init__.py
@@ -1,3 +1,4 @@
+from .base import AnalysisFunction  # noqa
 from .light_curve import LightCurve  # noqa
 from .stetsonj import *  # noqa
 from .structurefunction2 import *  # noqa
--- a/src/tape/analysis/base.py
+++ b/src/tape/analysis/base.py
@@ -1,0 +1,103 @@
+"""
+Contains the base class for analysis functions.
+"""
+
+from abc import ABC, abstractmethod
+from typing import List
+
+import pandas as pd
+
+
+class AnalysisFunction(ABC):
+    """Base class for analysis functions.
+
+    Analysis functions are functions that take few arrays representing
+    an object and return a single pandas.Series representing the result.
+
+    Methods
+    -------
+    cols(ens) -> List[str]
+        Return the columns that the analysis function takes as input.
+    meta(ens) -> pd.DataFrame
+        Return the metadata pandas.DataFrame required by Dask to pre-build
+        a computation graph. It is basically the schema for calculate() method
+        output.
+    on(ens) -> List[str]
+        Return the columns to group source table by.
+        Typically, `[ens._id_col]`.
+    calculate(*cols, **kwargs) -> pd.Series
+        Calculate the analysis function.
+    """
+
+    @abstractmethod
+    def cols(self, ens: "Ensemble") -> List[str]:
+        """
+        Return the column names that the analysis function takes as input.
+
+        Parameters
+        ----------
+        ens : Ensemble
+            The ensemble object, it could be required to get column names of
+            the "special" columns like `ens._time_col` or `ens._err_col`.
+
+        Returns
+        -------
+        List[str]
+            The column names to select and pass to .calculate() method.
+            For example `[ens._time_col, ens._flux_col]`.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def meta(self, ens: "Ensemble"):
+        """
+        Return the schema of the analysis function output.
+
+        Parameters
+        ----------
+        ens : Ensemble
+            The ensemble object.
+
+        Returns
+        -------
+        pd.DataFrame or (str, dtype) tuple or {str: dtype} dictionary
+            Dask meta, for example
+            `pd.DataFrame(columns=['x', 'y'], dtype=float)`.
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def on(self, ens: "Ensemble") -> List[str]:
+        """
+        Return the columns to group source table by.
+
+        Parameters
+        ----------
+        ens : Ensemble
+            The ensemble object.
+
+        Returns:
+        --------
+        List[str]
+            The column names to group by. Typically, `[ens._id_col]`.
+        """
+        return [ens._id_col]
+
+    @abstractmethod
+    def calculate(self, *cols, **kwargs):
+        """Calculate the analysis function.
+
+        Parameters
+        ----------
+        *cols : array_like
+            The columns to calculate the analysis function on. It must be
+            consistent with .cols(ens) output.
+        **kwargs
+            Additional keyword arguments.
+
+        Returns
+        -------
+        pd.Series or pd.DataFrame or array or value
+            The result, it must be consistent with .meta() output.
+        """
+        raise NotImplementedError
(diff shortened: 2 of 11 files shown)


dougbrn (changes requested):
This look great @hombit. My main request is to update the docs where a few lingering issues remain, particularly there's a few instances where it looks like calc_sf2 is called on array data, and those should be changed to calc_sf2.calculate calls.

hombit:
@dougbrn thank you! BTW, I'm not really happy about this `.calculate()` method: `calc_sf2.calculate` sounds weird. How do you think, could we replace it with `__call__()` or rename `calc_sf2 -> sf2` and `calc_stetson_J -> stetson_J`?

dougbrn:
> @dougbrn thank you! BTW, I'm not really happy about this `.calculate()` method: `calc_sf2.calculate` sounds weird. How do you think, could we replace it with `__call__()` or rename `calc_sf2 -> sf2` and `calc_stetson_J -> stetson_J`?

Hmm, I'm not sure that `__call__()` is ideal, as users may be annoyed/confused by the instantiation needed:
```
class AnalysisFunction:
    def __init__(self):
        print("Instance Created")

   def columns(self):
       return("flux")
  
    # Defining __call__ method
    def __call__(self, a, b):
        print(a * b)
  
# Instance created
ans = AnalysisFunction()
  
# __call__ method will be called
ans(10, 20)

# but we still can't just directly use this as a function...
AnalysisFunction(10,20)
```

Renaming the calc functions seems more reasonable, it looks like you currently have calc_stetson_J being imported as it's assigned as `calc_stetson_J = StetsonJ`, can we just let users directly import the StetsonJ class?

hombit:
I agree, with the only correction: currently `calc_stetson_J = StetsonJ()`, so `calc_stetson_J` is a callable object, an instance of `StetsonJ` class. Both things are importable, but I don't see how the class could be useful, because it has a trivial constructor.

dougbrn:
Alternatively, we could define the calculate function as a staticmethod of the class, like this:

```
class AnalysisFunction:
    def __init__(self):
        print("Instance Created")

    def columns(self):
       return("flux")
  
    # Defining __call__ method
    @staticmethod
    def analysisfunction(a, b):
        print(a * b)
  
# Call via Ensemble -- uses the class
ens.batch(AnalysisFunction)

# Call as function independently -- uses the staticmethod
AnalysisFunction.analysisfunction(10,20)
```

But this may be annoying to users as well that attempt to pass the static method to the ensemble, or try to act on the class like a function.

docs/tutorials/working_with_structure_function.ipynb
```diff
@@ -16,7 +16,7 @@
    "source": [
```
dougbrn, line 16: <p>Ah, this is tricky. Calculate should not be added here, as sf2 is a function of the timeseries class as invoked here. There's a few instances of that throughout this doc. It looks like timeseries.sf2 and ensemble.sf2 have also not been updated with the rename. As a future note, we may want to deprecate the unique sf2 functions in ensemble and timeseries in favor of a different batch function that serves a "glob" of lightcurves to a function.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/206/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

hombit: Thank you, I did too mechanically! I'll fix it

dougbrn (approved):
Nice, this looks good now with all the CI squared away!

hombit:
@dougbrn Should I improve test coverage?

dougbrn:
If you have ideas for further tests, feel free to add them. But I don't think that any of the loss in coverage here is especially impactful or merge-preventing

hombit:
I added a couple of new tests, but not going to add anything more

nevencaplar:
I talked with @hombit today; and I just reiterate that the main functionality that I want to have preserved is that an user can invoke our ``special functions'' just as `StetsonJ` or `sf2` without invoking the whole TAPE ensemble. I think we all are behind that and understand that. i know that is not the main point of this pull, but I find it so important that I am willing to repeat myself.

In best case scenario the flow is 
```
t = [1, 2, 3]
y = [2, 3, 4]
yerr = [3, 4, 5]
result = tape.sf2(t,y,yerr)
``` 

I understand that this pull changes this. I _think_ that I am fine with`result=tape.analysis.sf2.calculate(t, y, yerr)` and `result=tape.analysis.sf2.sf2(t,y,yerr)` . I think in tutorial we want to probably simplify that with invokation of `import tape.analysis.sf2 as sf2`.

hombit:
@dougbrn and @nevencaplar. I've just pushed the code to 1) rename functions back to `calc_*`, and 2) to rename `calculate()` to `__call__`, so finally this PR provides no interface change for analysis functions. If you are ok with this, I rebase and merge it

dougbrn:
I'm in favor of the change to preserve the current interface for analysis functions. I just have one last question for the ensemble.batch side, doesn't this implementation mean that a user should supply the StetsonJ class to batch, as supplying the calc_stetson_J instance will not be picked up by ensemble.batch's check to see if it's a subclass of AnalysisFunction? If this is the case, we should point this out in the documentation as much as possible, as I can see users easily just using the calc_stetson_J instance with the ensemble.

hombit:
Actually this PR changes nothing in how we use analysis functions with `batch`. New `batch` implementation checks if an input function is an instance of `AnalysisFunction`, and both `calc_sf2` and `calc_stetson_J` are instances of `AnalysisFunction`'s child classes:
```python
In [2]: from tape.analysis import calc_stetson_J, AnalysisFunction

In [3]: isinstance(calc_stetson_J, AnalysisFunction)
Out[3]: True
```

dougbrn:
Oh! I see, that's really cool.

Issue #211, opened 2023-08-29
Allow Ensemble with no Dask distributed client


hombit:
Currently, there is no simple way to create an Ensemble instance without an associated dask.distributed.Client. While having a client-less Ensemble may not be very useful from a scientific standpoint, it could make Ensemble more lightweight. See issues #208 and #199 for relevant discussions. I've also found that debugging errors with Client is much more difficult, and, for example, pytest-generated errors are much less readable. Therefore, we could probably use a client-less approach in most tests.

I think we can modify `Ensemble` constructor to accept `Client=False` or `="no"` or something like that.

nevencaplar:
In general I agree, but I am worried how technically feasible it is. Looking forward to input from Doug and Wilson.

dougbrn:
+1 to the notion that we should enable clientless setup. How we should represent the API for this is a bit tricky. To align with Dask, client should be able to accept a Dask.distributed.client object, currently if `client=None` then a client is spun up. The least invasive change could be to have client take in a Dask.distributed.client as a possible input, as before, or also accept a boolean, where `client=True` spins up a client and passes along kwargs, and `client=False` doesn't.

dougbrn:
In the above API, I would still advocate for client=True to be the default behavior

hombit:
@dougbrn I agree, this issue does not propose changing of the default behavior - just giving one more option

dougbrn:
Sounds good, I can try taking a stab at this.

hombit:
Fixed by https://github.com/lincc-frameworks/tape/pull/216

Issue #213, opened 2023-08-30
Convert docstring examples to doctests


hombit:
Currently "Example" sections in docstring are just raw text, not visible for pytest. I suggest converting them to actual doc-tests and run them with `pytest --doctest-modules` in CI. See numpy style guide here for details:
https://numpydoc.readthedocs.io/en/latest/format.html

PR #216, opened 2023-08-30
enable clientless ensemble

dougbrn:
Addresses #211. The unit test suite has been largely changed to not use a client, except for a few functions with additionally test the parquet_ensemble_with_client fixture. I have some concern about shifting many of our tests off of the distributed client, but it's possible that just using a subset of the test suite to cover distributed behavior in localized workflows will be sufficient. Since @hombit mentioned one of the pain points of the forced client was firewall permissions spam during the unit test suite execution, this implementation would allow for the small client subset to be excluding using this pytest incantation: `pytest -k "not with_client" `

I was originally trying to find a way to make a custom pytest flag that could switch the whole test suite from using the dask client or not, with the ambition to add client tests and clientless tests as two separate runs per python version in our github actions workflows. But I didn't have much luck parsing the pytest docs to understand how feasible this was.

### DIFF
--- a/docs/tutorials/scaling_to_large_data.ipynb
+++ b/docs/tutorials/scaling_to_large_data.ipynb
@@ -28,9 +28,9 @@
    "source": [
     "### The `Dask` Client and Scheduler\n",
     "\n",
-    "An important aspect of `Dask` to understand for optimizing it's performance for large datasets is the Client. `Dask` has [thorough documentation](https://distributed.dask.org/en/stable/client.html) on this, but the general idea is that the Client is an interface for sending instructions to a Scheduler, which in turn coordinates the worker nodes in executing the intended workflow. In essence, the Client gives you control over how to parallelize your workflow.\n",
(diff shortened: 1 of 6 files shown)


hombit:
Thanks, @dougbrn. I'll look more precisely a bit later, for now I just want to clarify that I'm fine about using a distributed client for most of the tests, because it would be our main and default use-case. No-client is useful if one of the tests fail and someone would like to have a simpler way to debug, or in the case of doc-tests (see #213) where it is not possible (??) to use pytest fixtures.

dougbrn:
Interesting, gotcha. Part of my rationale to switch most tests off of the client, is that the subset that remains all have "with_client", so you can do local runs without needing the client invoked anywhere. If that's less of a value add, then perhaps we can consider shifting most of the tests back onto the client?

hombit:
Yea, let's keep old tests to work as they used to

hombit (approved):
Looks great, thank you!

PR #218, opened 2023-08-31
Preserve EnsembleFrame metadata after assign()

wilsonbb:
A small fix for `EnsembleFrame.assign()` to correctly propagate the `label` and `ensemble` fields. This appears in calls to `EnsembleFrame.convert_flux_to_magnitude` but was overlooked in the initial testing. Fixed by adding `tape._Frame.assign` which just calls the parent and propagates metadata like we do in `tape._Frame.copy`


dougbrn (approved):
Looks good to me!

Issue #223, opened 2023-09-06
Construct Ensemble from Dask data-frames


hombit:
Following #221 I suggest adding a lower-level loading method `Ensemble.from_df()` accepting either a single source data-frame, or a pair of source and object data-frames. This would allow user loading data tables in a way they need - from CSV, FITS, etc, - and convert them to `Ensemble`.

I'm not sure about the Dask interface for data loading, but it would be perfect if we could support different data-frame objects, e.g. `dask.dataframe.DataFrame`, `pandas.DataFrame`, `astropy.table.Table`.

PR #224, opened 2023-09-08
Support loading ensemble from pandas and dask dataframes

wilsonbb:
Adds the loader functions `Ensemble.from_pandas` and `Ensemble.from_dask_dataframe`. This allows the user to provide TAPE with their own Pandas or Dask dataframes constructed via their own preferred sources and I/O as outlined in issue [#223](https://github.com/lincc-frameworks/tape/issues/223).

`Ensemble.from_source_dict` and `Ensemble.from_parquet` were simplified with some of their functionality moved into `Ensemble.from_dask_dataframe`.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -847,8 +847,7 @@ class Ensemble:
         """
         # Construct Dask DataFrames of the source and object tables
         source = dd.from_pandas(source_frame, npartitions=npartitions)
-        object = None if object_frame is None else dd.from_pandas(
-            object_frame, npartitions=npartitions)
+        object = None if object_frame is None else dd.from_pandas(object_frame, npartitions=npartitions)
         return self.from_dask_dataframe(
             source,
             object_frame=object,
@@ -856,7 +855,7 @@ class Ensemble:
             sync_tables=sync_tables,
             npartitions=npartitions,
             partition_size=partition_size,
-            **kwargs
+            **kwargs,
         )
 
     def from_dask_dataframe(
@@ -1131,15 +1130,12 @@ class Ensemble:
                     columns.append(col)
 
         # Read in the source parquet file(s)
-        source = dd.read_parquet(
-            source_file, index=self._id_col, columns=columns, split_row_groups=True
-        )
+        source = dd.read_parquet(source_file, index=self._id_col, columns=columns, split_row_groups=True)
 
         # Generate a provenance column if not provided
         if self._provenance_col is None:
             source["provenance"] = provenance_label
             self._provenance_col = "provenance"
-
 
         object = None
         if object_file:
@@ -1152,7 +1148,7 @@ class Ensemble:
             sync_tables=sync_tables,
             npartitions=npartitions,
             partition_size=partition_size,
-            **kwargs
+            **kwargs,
         )
 
     def from_dataset(self, dataset, **kwargs):
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -847,7 +847,7 @@ class Ensemble:
         """
         # Construct Dask DataFrames of the source and object tables
         source = dd.from_pandas(source_frame, npartitions=npartitions)
-        object = None if object_frame is None else dd.from_pandas(object_frame, npartitions=npartitions)
+        object = None if object_frame is None else dd.from_pandas(object_frame)
         return self.from_dask_dataframe(
             source,
             object_frame=object,
@@ -901,7 +901,7 @@ class Ensemble:
         # Set the index of the source frame and save the resulting table
         self._source = source_frame.set_index(self._id_col, drop=True)
 
-        if object_frame is None:  # generate object table from source
+        if object_frame is None:  # generate an indexed object table from source
             self._object = self._generate_object_table()
             self._nobs_bands = [col for col in list(self._object.columns) if col != self._nobs_tot_col]
         else:
@@ -917,6 +917,8 @@ class Ensemble:
             if self._nobs_tot_col is None:
                 self._object["nobs_total"] = np.nan
                 self._nobs_tot_col = "nobs_total"
+
+            self._object.set_index(self._id_col)
 
             # Optionally sync the tables, recalculates nobs columns
             if sync_tables:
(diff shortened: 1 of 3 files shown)


wilsonbb:
Will add an additional test to bump up the code coverage, but I wanted it to be ready for a first review pass.

tests/tape_tests/conftest.py
```diff
@@ -118,3 +121,58 @@ def parquet_ensemble_from_hipscat(dask_client):
...
+# pylint: disable=redefined-outer-name
+@pytest.fixture
+def pandas_ensemble(dask_client):
+    """Create an Ensemble from parquet data."""
+    ens = Ensemble(client=dask_client)
+
+    num_points = 1000
+    all_bands = ["r", "g", "b", "i"]
+    rows = {
+        "id": [8000 + (i % 5) for i in range(num_points)],
```
hombit, line 162: Maybe not important in tests, but list comprehension is slow, I would prefer to write
```python
8000 + (np.arange(num_points) % 5)
```
or even
```python
8000 + np.tile(np.arange(num_objects), num_points // num_objects)
```
where `num_objects` is 5

src/tape/ensemble.py
```diff
@@ -1002,49 +1126,26 @@ def from_parquet(
...
-            self._object = self._generate_object_table()
-            self._nobs_bands = [col for col in list(self._object.columns) if col != self._nobs_tot_col]
+        source = dd.read_parquet(source_file, index=self._id_col, columns=columns, split_row_groups=True)
 
         # Generate a provenance column if not provided
         if self._provenance_col is None:
-            self._source["provenance"] = self._source.apply(
-                lambda x: provenance_label, axis=1, meta=pd.Series(name="provenance", dtype=str)
-            )
+            source["provenance"] = provenance_label
```
hombit, line 1133: What is provenance, do we need it for pandas / dask dataframe inputs?

src/tape/ensemble.py
```diff
@@ -1002,49 +1126,26 @@ def from_parquet(
...
+            source_frame=source,
+            object_frame=object,
+            column_mapper=column_mapper,
+            sync_tables=sync_tables,
+            npartitions=npartitions,
+            partition_size=partition_size,
+            **kwargs,
+        )
 
     def from_dataset(self, dataset, **kwargs):
```
hombit, line 1150: I love how it becomes much shorter!

hombit (approved):
Thank you, it looks great! I list some general questions and non-important code suggestions.

src/tape/ensemble.py
wilsonbb: In the case of multi-survey data, the "provenance" column is meant to determine which rows correspond to which survey.

My thoughts are that the pandas/dask dataframe case should be close to the `from_source_dict` where we currently don't populate this for the user. Being a lower-level interface users can always populate the field themselves if needed.

tests/tape_tests/conftest.py
wilsonbb: Thanks for suggesting! Updated the other list comprehensions as well.

tests/tape_tests/test_ensemble.py
```diff
@@ -67,6 +67,42 @@ def test_from_parquet(data_fixture, request):
...
+    assert ens._source is not None
+    assert ens._object is not None
+
+    # Check that the data is not empty.
+    obj, source = ens.compute()
+    assert len(source) == 1000
+    assert len(obj) == 5
+
+    # Check that source and object both have the same ids present
+    assert sorted(np.unique(list(source.index))) == sorted(np.array(obj.index))
```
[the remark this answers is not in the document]
wilsonbb: Done.

src/tape/ensemble.py
```diff
@@ -807,6 +807,130 @@ def s2n_inter_quartile_range(flux, err):
...
+            of size `partition_size`.
+
+        Returns
+        ----------
+        ensemble: `tape.ensemble.Ensemble`
+            The ensemble object with the Dask dataframe data loaded.
+        """
+        # Construct Dask DataFrames of the source and object tables
+        source = dd.from_pandas(source_frame, npartitions=npartitions)
+        object = None if object_frame is None else dd.from_pandas(object_frame, npartitions=npartitions)
```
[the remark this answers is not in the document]
wilsonbb: These are good questions!

1. If I'm understanding correctly, in cases where we can fit our data in a pandas dataframe in RAM, we honestly probably don't gain anything from using Dask (see [Dask best practices](https://docs.dask.org/en/stable/dataframe-best-practices.html#use-pandas)). The `from_pandas` and `from_source_dict` cases are more for allowing a user to play around with smaller data sets, but I don't think they'll be very performant. I think if we want to support those use cases more, we'll probably want the `source` and `object` frames to be in pandas rather than Dask. @dougbrn might have opinions here when he gets back.

2. A call to `dask.dataframe.repartitions` where `npartitions` is the frame's current number of partitions is a no-op (see [here](https://github.com/dask/dask/blob/main/dask/dataframe/core.py#L7854C38-L7854C38)). So my thoughts were to go ahead and specify the number of partitions at frame creation in the hopes that we might save the cost of partitioning twice. Calling repartition in `from_dask_dataframe` of course just gives the user an easy way to repartition their frame if needed.

However in `from_parquet` we don't specify the number of partitions in the object table. So removing specifying npartitions for the object table here to conform to that behavior.

3. Yes having a partitioning scheme to avoid/limit this is important. Note that the source frame's index is based on the object ID, so dask's partitioning scheme should be trying to group together source data by object. But it might turn out we need to add some more intelligent ways to prevent this. I believe you are correct about nodes and partitions.

4. Yeah, I'm worried about this as well. The fact that both frames should be sorted by the same object ID index should allow Dask to try to avoid/limit cases like these (see about sorted joins [here](https://docs.dask.org/en/stable/dataframe-joins.html#sorted-joins)), but I think we'll need to see what the costs are in practice. I've modified `from_dask_dataframe` to ensure the object table is always indexed by the object ID regardless of how the user created the object frame.

And yeah if you feel like you want to followup over Slack rather than here, feel free to reach out and I can then summarize the discussion here :)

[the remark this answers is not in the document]
hombit: Thank you!

Issue #230, opened 2023-09-14
Ensemble.from_dask_dataframe set the index wrongly


hombit:
I believe that `Ensemble.from_dask_dataframe()` does something wrong while setting indexes if they are not-pre set by the user to be `_id_col` (and not anything else)

```python
import dask.dataframe as dd
import numpy as np
from tape import Ensemble, ColumnMapper, calc_stetson_J

n_obj = 10
n_src_per_obj = 20
id = np.arange(10)
name = id.astype(str)
object_table = dd.from_dict(
    dict(id=id, name=name),
    npartitions=1,
)

time = np.repeat(np.linspace(0, 1, n_src_per_obj), n_obj)
source_table = dd.from_dict(
    dict(
        id=np.tile(id, n_src_per_obj),
        time=time,
        flux=time,
        flux_err=np.full_like(time, 1e-3),
        band=np.full(time.shape, "g"),
    ),
    npartitions=1,
)

ens = Ensemble(client=False)
ens.from_dask_dataframe(
    source_frame=source_table,
    object_frame=object_table,
    npartitions=1,
    column_mapper=ColumnMapper(
        id_col='id',
        time_col='time',
        flux_col='flux',
        err_col='flux_err',
        band_col='band',
    ),
)

amplitude = ens.batch(calc_stetson_J)
```

```
ValueError: 'id' is both an index level and a column label, which is ambiguous.
```

It fixes if I do `set_index` for both tables before building an ensemble

PR #231, opened 2023-09-14
Set object table index for from_dask_dataframe

wilsonbb:
In the loader method `from_dask_dataframe`, we were mistakenly not updating the object table's index in the case where the object table is supplied as an argument rather than generated from the source table.

Unit tests are updated based on the identifying issue [#230](https://github.com/lincc-frameworks/tape/issues/230) validating the case where the object table isn't generated and that analysis functions can be called on the generated  ensemble.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -852,7 +852,7 @@ class Ensemble:
         """
         # Construct Dask DataFrames of the source and object tables
         source = dd.from_pandas(source_frame, npartitions=npartitions)
-        object = None if object_frame is None else dd.from_pandas(object_frame)
+        object = None if object_frame is None else dd.from_pandas(object_frame, npartitions=npartitions)
         return self.from_dask_dataframe(
             source,
             object_frame=object,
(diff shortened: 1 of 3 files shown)


hombit (approved):
Looks great! The only minor issue is a warning in the case of pre-existing index
```
dask/dataframe/core.py:5251: UserWarning: New index has same name as existing, this is a no-op.
  warnings.warn(
```

dougbrn (approved):
Looks good!

wilsonbb:
> Looks great! The only minor issue is a warning in the case of pre-existing index
> 
> ```
> dask/dataframe/core.py:5251: UserWarning: New index has same name as existing, this is a no-op.
>   warnings.warn(
> ```

Considering this some more, I've decided to open an issue to clarify documentation or provide a parameter to allow users to force the indices to be overridden: https://github.com/lincc-frameworks/tape/issues/232

However I will go ahead and submit this PR as is while I think about the interface a bit more.

Thanks for the reviews!

PR #234, opened 2023-09-15
Adds ability to give a single float input to zero_point

dougbrn:
Motivated by Pan-STARRS1 zero_point calculations. Currently, the zero_point needs to be a column of the source table. But it seems clear from some limited examples that users will often want to just provide a float here.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1247,9 +1247,10 @@ class Ensemble:
         ----------
         flux_col: 'str'
             The name of the ensemble flux column to convert into magnitudes.
-        zero_point: 'str'
+        zero_point: 'str' or 'float'
             The name of the ensemble column containing the zero point
-            information for column transformation.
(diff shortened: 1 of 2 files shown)


src/tape/ensemble.py
```diff
@@ -1247,9 +1247,10 @@ def convert_flux_to_mag(self, flux_col, zero_point, err_col=None, zp_form="mag",
         ----------
         flux_col: 'str'
             The name of the ensemble flux column to convert into magnitudes.
-        zero_point: 'str'
+        zero_point: 'str' or 'float'
             The name of the ensemble column containing the zero point
-            information for column transformation.
+            information for column transformation. Alternatively, a float zero
```
hombit, line 1252: It may be may issue, but `float zero point` sounds a bit weird. Could we say something like
"The name of the ensemble column containing the zero point values. Alternatively, a single float number to apply for all fluxes."

hombit (approved):
Thank you!

PR #237, opened 2023-09-19
Fix EnsembleFrame.set_index

wilsonbb:
Fixes issue where `EnsembleFrame.set_index` does not correctly set a named index.

As an example, `EnsembleFrame.set_index("id")` would fail to produce an `EnsembleFrame` with an index named "id". This would not immediately fail, but performing `groupby["id"]` on the resulting frame would result in a key error since we dropped the column and our index is unnamed. Such a test case has now been added in this PR.

In the previous behavior, the calls to `set_index` eventually produce a call to `dask.dataframe.core.map_partitions` and the incorrect index is constructed in [`_get_meta_map_partition`](https://github.com/dask/dask/blob/216c7c9393d8df1014a911eceb797502999e8a51/dask/dataframe/core.py#L7066). Here the meta is overridden with an incorrect index in the call to

```
meta_index = getattr(make_meta(dfs[0]), "index", None) if dfs else None

...

meta = make_meta(meta, index=meta_index, parent_meta=parent_meta)
```

where the `meta_index` is the unnamed index that was causing the error described above. After some investigation, I discovered that even in the case of a normal dask.DataFrame, `meta_index` is also an unnamed index, but the call to `make_meta` was not causing it to override the existing index. 

A fix is provided in this PR by simply having our calls to `make_meta_*` not override the existing indices.

### DIFF
--- a/src/tape/ensemble_frame.py
+++ b/src/tape/ensemble_frame.py
@@ -1,3 +1,5 @@
+from collections.abc import Sequence
+
 import dask.dataframe as dd
 
 import dask
@@ -9,6 +11,9 @@ from dask.dataframe.extensions import make_array_nonempty
 
 import numpy as np
 import pandas as pd
+
+from typing import Literal
+
 
 from functools import partial
 from dask.dataframe.io.parquet.arrow import (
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
Nice, this looks good to me!

PR #240, opened 2023-09-28
Update requirements.txt to unpin sphinx

OliviaLynn:
Noticed this while working on my dash; unpinning sphinx to address doc build failure (see https://github.com/lincc-frameworks/python-project-template/issues/294)


dougbrn (approved):
Thanks for finding this, looks good!

Issue #246, opened 2023-09-28
Add explicit nobs calculator functions


dougbrn:
Related to #243 and #245

We should give users a function to explicitly calculate nobs_total and nobs by band information, as would be especially needed in the "lean" mode case mentioned in #245

dougbrn:
Will take this as a pre-cursor to #245

dougbrn:
Closed by #251

PR #251, opened 2023-10-02
add calc_nobs

dougbrn:
Adds a function to calculate the number of observations in source, with option to do it just for the total, or the total and a count for each band.

Addresses #246.


wilsonbb:
Looks good to me!

wilsonbb (approved):

PR #252, opened 2023-10-03
Add update_ensemble() and Use EnsembleFrames

wilsonbb:
Adds support for `EnsembleFrame.update_ensemble()` and uses the `ObjectFrame` and `SourceFrame` objects throughout the ensemble. 

Note that in order for `_generate_object_table()` to work we also needed to add `ObjectFrame.from_dask_dataframe` to create an `ObjectFrame` from the generated dask table. 

We also move the fields for whether an object or source table is "dirty" into fields within the EnsembleFrames themselves.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -133,7 +133,8 @@ class Ensemble:
 
         Raises
         ------
-        ValueError if the `frame.label` is unpopulated, "source", or "object".
+        ValueError if the `frame.label` is unpopulated, or if the frame is not a SourceFrame or ObjectFrame
+        but uses the reserved labels.
         """
         if frame.label is None:
             raise ValueError(
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -12,7 +12,7 @@ from .analysis.base import AnalysisFunction
 from .analysis.feature_extractor import BaseLightCurveFeature, FeatureExtractor
 from .analysis.structure_function import SF_METHODS
 from .analysis.structurefunction2 import calc_sf2
-from .ensemble_frame import ObjectFrame, SourceFrame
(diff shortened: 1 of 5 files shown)


wilsonbb:
There are still some errors with batch sometimes returning an empty object tbale, but I thought I would go ahead and send this part out for review.

src/tape/ensemble.py
```diff
@@ -1160,14 +1163,13 @@ def from_parquet(
                     columns.append(col)
 
         # Read in the source parquet file(s)
-        self._source = dd.read_parquet(
-            source_file, index=self._id_col, columns=columns, split_row_groups=True
```
dougbrn, line 1164: just curious, why are we dropping the split_row_groups kwarg here and below?

src/tape/ensemble.py
```diff
@@ -1182,12 +1184,12 @@ def from_parquet(
 
             # Optionally sync the tables, recalculates nobs columns
             if sync_tables:
-                self._source_dirty = True
-                self._object_dirty = True
+                self._source.set_dirty(True)
```
dougbrn, line 1187: Nice, I like the dirty flags as a property of the frame itself.

dougbrn (approved):
This looks good to me, just had a small non-blocking question!

src/tape/ensemble.py
wilsonbb: Yeah, a good question! This is just because in [`EnsembleFrame.from_parquet`](https://github.com/lincc-frameworks/tape/blob/main/src/tape/ensemble.py#L1180) we currently always have `split_row_groups=True`.

This was meant to match that we previously never let a user set to false, but of course we could open up the option (and then specify it here).

Issue #245, opened 2023-09-28
Improvements to Table Syncing


dougbrn:
Related to #243 

Table Syncs are a key component of the TAPE Ensemble design, where Object and Source are kept up to date with one another as the user operates on one or both tables. Currently with Table Syncs, we do the following:
* Object to Source Syncs: When Object is modified, grab the index of object and filter Source based on that index
* Source to Object Syncs: When Source is modified, groupby each (object, band) pairing, count them to produce a skinny object table with nobs_total, and an nobs column for each band populated. Join the current object table to the new object table.

With the Source to Object Sync, we are doing much more work to perform the sync, because we assume that these nobs columns will be useful to users. However, from discussion and some real workflows, it's clear that some users will not need these nobs columns for their work. In these cases, we are introducing needless operations to the workflow.

Given this, we should introduce more options for ensemble syncing. 
* At minimum, users should be able to have a "lean" sync mode, that only drops rows from object/source based on lack of occurence of ids in the other table (except in the case where the user has set the flag to have Object preserve ids with no source information).
* Another available option should be to track just nobs_total for source to object syncs
* And, dependent on #243, have the option to sync the full nobs by band as is currently done by TAPE

Beyond this, nobs information is likely not the only information that a user may want to sync. As @hombit mentioned: duration (per band), mean mag, variability index, etc. are all examples of potentially trackable metrics. There is potential to abstract syncs to enable users to setup their own sync criteria, something where given a set of input columns from one table, a function, and a set of output columns TAPE is instructed to run that function whenever the dirty flag is set. We should make sure that any implementation has use cases demonstrating clear upside to the alternative, of users just manually running functions to update columns when they need that column for the next analysis step.

dougbrn:
Additional points from meeting discussion:
* I still don't neccesarily think we have strong use cases where sync generating any additional information is helpful to users, beyond just having an explicit function to generate that information when the user needs it.
* A downside to the explicit function route, is that these results columns can become out of date. For example, if I calculate nobs_total, then filter the source table, nobs_total is no longer correct.
* In the above case, one potential consensus was that it would be good to flag the nobs_total column when it's generated as a column that can become out of date ("transient" column? Degeneracy with astro lingo...). When the parent table of a "transient" column is dirty, the sync should involve dropping that column. Motivation being it's better for users to fail to grab bad information, than be able to operate on bad information.
[...]

PR #254, opened 2023-10-04
Reduce scope of sync_tables

dougbrn:
Addresses some first steps of #245 

* Removes nobs columns from critical columns, no longer generates them from source in `_generate_object_table()`
* Sync_tables now just aligns indices, no generation of nobs information
* Adds the notion of temporary columns, which are removed when syncs occur
* Updates prune() to use an existing nobs column if provided, but otherwise calls `calc_nobs()` as part of the operation


src/tape/ensemble.py
```diff
@@ -538,19 +541,38 @@ def calc_nobs(self, by_band=False, label="nobs"):
                 .reset_index()  # break up the multiindex
                 .categorize(columns=[self._band_col])  # retype the band labels as categories
                 .pivot_table(values=self._band_col, index=self._id_col, columns=self._band_col, aggfunc="sum")
-                .repartition(obj_npartitions)  # counts inherits the source partitions
             )  # the pivot_table call makes each band_count a column of the id_col row
 
+            # repartition the result to align with object
+            if self._object.known_divisions:
+                band_counts = band_counts.reset_index().set_index(self._id_col)  # ugly, but need this
```
dougbrn, line 548: I had to do this to resolve issues with assign not knowing divisions, and I'm really not happy about it. If anyone has ideas on how to more elegantly generate division information, let me know.

src/tape/ensemble.py
hombit, line 1: Could we also have `temporary` argument for assign? So users could add their own temporary columns.

hombit (approved):
Looks good, thank you!

src/tape/ensemble.py
dougbrn: Yeah I think this makes a lot of sense!

dougbrn:
@hombit I can't seem to reply in line with the question re: why the reset_index.set_index is needed. The problem in some cases seems to be that the grouping is done on the source table, so the source partitions are preserved. And the divisions are not known unless they were known for source. The assign call seems to have issues when the divisions for object are known, but the divisions for the new pandas series to assign are not. That being said, I wonder if it'd resolve better if I tried to remove the division information from object, might give that a go

src/tape/ensemble.py
dougbrn: The grouping is done on the source table, so the source partitions are preserved. And the divisions for the resulting dataframe are not known unless they were known for source. This leads to the problem where the assign call seems to have issues when the divisions for object are known, but the divisions for the new pandas series to assign are not. That being said, I wonder if it'd resolve better if I tried to remove the division information from object, might give that a go

src/tape/ensemble.py
```diff
@@ -538,19 +541,38 @@ def calc_nobs(self, by_band=False, label="nobs"):
...
+                band_counts = band_counts.repartition(npartitions=self._object.npartitions)
+
             # short-hand for calculating nobs_total
             band_counts["total"] = band_counts[list(band_counts.columns)].sum(axis=1)
 
             bands = band_counts.columns.values
             self._object = self._object.assign(**{label + "_" + band: band_counts[band] for band in bands})
 
+            if temporary:
+                self._object_temp.extend([label + "_" + band for band in bands])
```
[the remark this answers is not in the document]
dougbrn: I think this is actually needed as a generator for a list comprehension

src/tape/ensemble.py
dougbrn: The latest commit changes this to just drop divisions from object so it doesn't try to use them as part of the assign

src/tape/ensemble.py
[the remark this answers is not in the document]
hombit: I don't think so.

Practice:
```python
In [1]: a = [-1]
In [2]: a.extend(i**2 for i in range(5))
In [3]: a
Out[3]: [-1, 0, 1, 4, 9, 16]
```

Theory:
- `list.extend()` takes any iterable, not necessary list or even a collection. That means it is not nessery to create a new list, it is enough to pass a generator
- In-line generator syntax is `(expr for item in iter)`, with round brackets. However, if in-line generator is an only argument of a function, it is allowed to omit these brackets.

I think that not having square brackets is a more pythonic way of doing the things, but of course it is a matter of taste

PR #256, opened 2023-10-04
Issue 199: Added static Ensemble read constructors to tape namespace

wenneman:
Added 6 new read_ constructors which currently wrap the existing from_ data populating methods.


src/tape/ensemble.py
delucchi-cmu, line 1: This file is gigantic. We should think about alternate code organization.

Looking at the issue, I think the hope was just that these loader methods would be in the `tape` namespace, and not necessarily in the `tape.ensemble` namespace. I think we could get away with 

/src/tape/__init__.py
```
import (list 'em all out) from ensemble_readers
```

then move these new loading methods to `ensemble_readers.py`. These methods aren't using any private/protected methods on the ensemble to necessitate living right next to them.

src/tape/ensemble.py
```diff
@@ -1587,3 +1587,326 @@ def sf2(self, sf_method="basic", argument_container=None, use_map=True):
...
             return result
+
+
+"""
+    The following package-level methods can be used to create a new Ensemble object 
+    by reading in the given data source.
+"""
+
+
+def read_pandas_dataframe(
```
delucchi-cmu, line 1598: "read" doesn't seem like the right word, since you've already read the originating file, and you're just constructing the ensemble from already-in-memory pandas. `from_` feels like the best prefix, but that's kind of taken already on the ensemble itself. Maybe `load_`?

src/tape/ensemble.py
```diff
@@ -1587,3 +1587,326 @@ def sf2(self, sf_method="basic", argument_container=None, use_map=True):
...
+    Parameters
+    ----------
+    source_frame: 'pandas.Dataframe'
+        A Dask dataframe that contains source information to be read into the ensemble
+    object_frame: 'pandas.Dataframe', optional
+        If not specified, the object frame is generated from the source frame
+    dask_client: `dask.distributed.client` or `bool`, optional
+        Accepts an existing `dask.distributed.Client`, or creates one if
+        `client=True`, passing any additional kwargs to a
+            dask.distributed.Client constructor call. If `client=False`, the
```
delucchi-cmu, line 1619: Extra indentation. Carries through to a weird rendering in the API doc:

https://tape--256.org.readthedocs.build/en/256/autoapi/tape/ensemble/index.html#tape.ensemble.read_pandas_dataframe

Same thing in the subsequent methods.

src/tape/ensemble.py
```diff
@@ -1587,3 +1587,326 @@ def sf2(self, sf_method="basic", argument_container=None, use_map=True):
...
+
+def read_hipscat(
+    dir,
+    source_subdir="source",
+    object_subdir="object",
+    column_mapper=None,
+    dask_client=True,
+    **kwargs,
+):
+    """Read in parquet files from a hipscat-formatted directory structure
```
delucchi-cmu, line 1795:
```suggestion
    """Read in parquet files from a hipscat-formatted directory structure
    
```

delucchi-cmu: Missing newline

tests/tape_tests/conftest.py
```diff
@@ -2,14 +2,198 @@
...
+    n_obj = 5
+    id = 8000 + np.arange(n_obj)
+    name = id.astype(str)
+    return dd.from_dict(
+        dict(id=id, name=name),
+        npartitions)
+
+
+# pylint: disable=redefined-outer-name
+@pytest.fixture
```
delucchi-cmu, line 52: Do you expect to want to re-use these fixtures outside of `test_ensemble.py`? I usually think of the fixtures within the `conftest.py` as ones that you want to share extensively between different test modules. I think in this case, you can define the fixtures within `test_ensemble.py` directly, just to avoid bloating this shared file.

tests/tape_tests/conftest.py
```diff
@@ -2,14 +2,198 @@
...
 import pytest
+import tape
+
 from dask.distributed import Client
 
 from tape import Ensemble
 from tape.utils import ColumnMapper
 
 
+def create_test_rows():
```
delucchi-cmu, line 14: You could also make *this* a fixture, and create other fixtures with this fixture as input. Something like:

```
@pytest.fixture
def test_source_rows():
  ...

@pytest.fixture
def test_source_table(test_source_rows):
  ...

@pytest.fixture
def read_dask_dataframe_ensemble(dask_client: Client, test_source_table):
    return tape.read_dask_dataframe(
        dask_client = dask_client,
        source_frame = test_source_table,
        ...
    )
```

tests/tape_tests/test_ensemble.py
```diff
@@ -168,6 +179,50 @@ def test_from_qso_dataset(dask_client):
...
 
 
+def test_read_rrl_dataset(dask_client):
+    """
+    Test a basic load and analyze workflow from the S82 RR Lyrae Dataset
+    """
+
+    ens = tape.read_dataset("s82_rrlyrae", dask_client=dask_client)
+
+    # larger dataset, let's just use a subset of ~100
```
delucchi-cmu, line 189: Is it 100 or 350?

tests/tape_tests/test_ensemble.py
```diff
@@ -208,6 +263,52 @@ def test_from_source_dict(dask_client):
     assert obj_table.iloc[1][ens._nobs_tot_col] == 5
 
 
+def test_read_source_dict(dask_client):
+    """
+    Test that ensemble.from_source_dict() successfully creates data from a dictionary.
```
delucchi-cmu, line 268:
```suggestion
    Test that tape.read_source_dict() successfully creates data from a dictionary.
```

tests/tape_tests/test_ensemble.py
```diff
@@ -208,6 +263,52 @@ def test_from_source_dict(dask_client):
...
+    """
+    ens = Ensemble(client=dask_client)
```
delucchi-cmu, line 270:
```suggestion

```

src/tape/ensemble.py
wenneman: Good point. These methods were originally marked as @classmethod in the Ensemble object, but when I moved them out of the Ensemble namespace I neglected to move them out of that package/file. 

I'll correct that and place these methods in an `ensemble_readers.py` file as suggested, and still in the tape namespace.

src/tape/ensemble.py
wenneman: "read" is the naming convention used by Pandas and subsequently by Dask for these type of constructor operations. Hence followed that naming convention for tape.

https://pandas.pydata.org/docs/user_guide/io.html
https://docs.dask.org/en/latest/dataframe-api.html#create-dataframes

src/tape/ensemble.py
wenneman: Fixed (and surprised `black` didn't reformat these)

tests/tape_tests/conftest.py
wenneman: Thanks, that's what I was originally going for but wasn't sure how to do it. I've made that change now so that those helper functions are now a @pytest.fixture.

tests/tape_tests/test_ensemble.py
wenneman: Looks like this was a repeated copy and paste error. I've cleaned those up (and removed the comment reference to the subset size :).

tests/tape_tests/conftest.py
wenneman: New issue https://github.com/lincc-frameworks/tape/issues/257

tests/tape_tests/conftest.py
```diff
@@ -28,75 +28,89 @@ def create_test_rows():
...
+def create_test_column_mapper():
     return ColumnMapper(id_col="id", time_col="time", flux_col="flux", err_col="err", band_col="band")
 
-def create_test_source_table(npartitions=1):
 
-    return dd.from_dict(
-        create_test_rows(), 
-        npartitions)
+@pytest.fixture
+@pytest.mark.parametrize("create_test_rows", [("create_test_rows")])
```
delucchi-cmu, line 38: The `@pytest.mark.parametrize` feels extraneous here, but I'm guessing you did what you had to do to get the tests to work =]

delucchi-cmu (approved):

PR #262, opened 2023-10-06
fix bug in sync

dougbrn:
Small bug fix for the sync PR that just went in.


drewoldag (approved):
Looks good. Any unit tests for this?

dougbrn:
@drewoldag There is an existing test that exercises this code, but it has a blindspot where it doesn't explicitly remove a whole objects worth of sources from the source table. This led to the unit test failing to actually catch when the object table wasn't syncing to source. I've added another commit which expands the test to catch this.

PR #264, opened 2023-10-06
Propagate EnsembleFrame._is_dirty

wilsonbb:
When transformations on an `EnsembleFrame` create a new object, its `_is_dirty` status isn't copied to the new object. We move the field into its parent `_Frame` class so it can be copied in `_Frame._propagate_metadata`

Also ensures that we track `EnsembleFrame` metadata (label, parent ensemble, and now `_is_dirty`) across merge, drop, and persist calls.

### DIFF
--- a/src/tape/ensemble_frame.py
+++ b/src/tape/ensemble_frame.py
@@ -202,20 +202,181 @@ class _Frame(dd.core._Frame):
         shuffle=None,
         broadcast=None,
     ):
+        """Merge the Dataframe with another DataFrame
+
+        Doc string below derived from dask.dataframe.core
+
+        This will merge the two datasets, either on the indices, a certain column
+        in each dataset or the index in one dataset and the column in another.
+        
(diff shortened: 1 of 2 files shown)


src/tape/ensemble_frame.py
```diff
@@ -178,6 +187,198 @@ def query(self, expr, **kwargs):
         result = super().query(expr, **kwargs)
         return self._propagate_metadata(result)
     
+    def merge(
```
dougbrn, line 190: Is explicitly listing the kwargs here needed, or can they just be passed along via **kwargs to dask's merge call?

dougbrn (approved):
This looks good, just had one question that I was curious about.

src/tape/ensemble_frame.py
wilsonbb: Yeah this seems the correct call. Thanks!

PR #269, opened 2023-10-10
Remove calls to set_dirty in ensemble

wilsonbb:
Throughout the Ensemble, we manually mark frames as dirty for operations that we expect to change the number of rows such as dropna. In PR https://github.com/lincc-frameworks/tape/pull/267, we Ensemble.update_frame perform this whenever the Ensemble has been updated with a frame with a different number of rows. This allows us to remove manual calls to set_dirty throughout the Ensemble since operations like dropna use Ensemble.update_frame to store their results.

Note that the previous sync tests called dropna despite no NaNs being present in the table. Since now dropna only marks the table as dirty when rows are actually dropped, the test tables were modified to include a NaN value. Otherwise the testing strategy is to just rely on existing tests.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -327,7 +327,6 @@ class Ensemble:
 
         # Append the new rows to the correct divisions.
         self.update_frame(dd.concat([self._source, df2], axis=0, interleave_partitions=True))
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
looks good!

Issue #275, opened 2023-10-23
Add Check functions for Sorting and Lightcurve "cohesion"


dougbrn:
For a given input dataset, it's status as being sorted on a given ID and whether or not the source information is split across partitions are important things to know for assessing performance, optimization, and even knowing when you're able to use kwargs like `use_map` within batch. I think it would be good to add two new functions that simply check the status of these two properties, as follows:

* Ensemble.check_sorted: Checks to see if the current ensemble (source and/or object) is sorted by the given id column
* Ensemble.check_lightcurve_cohesion: Checks to see if all sources for a given ID can only be found in a single partition, in other ways, should check that all partitions have a unique set of object IDs present.

Issue #242, opened 2023-09-28
set_index calls quietly sort the Ensemble dataframes


dougbrn:
Discovered this recently that the default behavior of `dd.set_index` is to sort the dataframe based on the index values. This introduces a costly overhead to the workflow, and is needless when the user already has their data sorted. In principle, TAPE should be able to function without a sorted index. We should consider how to best implement this sorting functionality as an optional feature, and give users the ability to not do it for datasets that don't require it.

Additional note, if we are sorting the dataframes, it may be worthwhile to investigate generation of division information.

hombit:
> We should consider how to best implement this sorting functionality as an optional feature, and give users the ability to not do it for datasets that don't require it.
The third option, which could be a default behavior, is checking the index to be sorted while reading the data. This is significantly cheaper than force-sorting, and would inform user if data must be sorted.

dougbrn:
Yeah it would be nice if we could scan and check for that. I'm not sure how costly it will be since Dask will have to actually verify that it is sorted, and that is an operation that can't be done lazily.

dougbrn:
Going to close this as the original scope of the issue has been addressed.

PR #276, opened 2023-10-23
add check functions

dougbrn:
This PR addresses #275, adding two functions (`check_sorted` and `check_lightcurve_cohesion`) that assess whether the ensemble data is sorted and whether the source data is co-located within single partitions (referring to this as "cohesion"). 

There's a bit of a snowball effect to implementing these, so this ends up addressing #242 by adding a `sort` flag to loader calls that defaults to False. This allows the Ensemble to actually be unsorted, making the `check_sorted` function useful. The sort flag was added to Dask in 2023.6.1, so I've also updated the minimum version to this in `pyproject.toml`, this broke something in `ensemble.insert_sources`, where a small tweak is needed to the ColumnMapper and sorting needs to be on for it and potentially any upstream data loading. This may be confusing to users, but I also see this function as mainly for devs and is probably unlikely that users will need it much?

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -152,7 +152,7 @@ class Ensemble:
 
         # Create the new row and set the paritioning to match the original dataframe.
         df2 = dd.DataFrame.from_dict(rows, npartitions=1)
-        df2 = df2.set_index(self._id_col, drop=True, sort=False)
+        df2 = df2.set_index(self._id_col, drop=True, sort=True)
 
         # Save the divisions and number of partitions.
         prev_div = self._source.divisions
@@ -169,6 +169,8 @@ class Ensemble:
                 self._source = self._source.repartition(divisions=prev_div)
             elif self._source.npartitions != prev_num:
                 self._source = self._source.repartition(npartitions=prev_num)
+
+        return self
 
     def client_info(self):
         """Calls the Dask Client, which returns cluster information
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -233,7 +233,7 @@ class Ensemble:
         # Use the existing index function to check if it's sorted (increasing)
         return idx.is_monotonic_increasing.compute()
 
-    def check_lightcurve_cohesion(ens):
+    def check_lightcurve_cohesion(self):
         """Checks to see if lightcurves are split across multiple partitions.
 
         With partitioned data, and source information represented by rows, it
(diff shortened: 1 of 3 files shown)


dougbrn:
This PR adds 2 lines that are uncovered by unit tests, its a conditional to check and raise an error if the user has supplied something other than "object" or "source" to `check_sorted`. I would argue this is fine, as these conditionals are about to go away across the Ensemble once the refactor goes in.

src/tape/ensemble.py
```diff
@@ -206,6 +209,54 @@ def info(self, verbose=True, memory_usage=True, **kwargs):
...
+        A boolean value indicating whether the index is sorted (True)
+        or not (False)
+        """
+        if table == "object":
+            idx = self._object.index
+        elif table == "source":
+            idx = self._source.index
+        else:
+            raise ValueError(f"{table} is not one of 'object' or 'source'")
+        return idx.map_partitions(lambda a: np.all(a[:-1] <= a[1:])).compute().all()
```
jeremykubica, line 231: It is not obvious to me how this works. Doesn't map_partitions call the function (in this case the lambda) once per partition? If so, it is only checking that the last element is smaller or equal to the first one. Wouldn't that leave out the elements in the middle?

From the tests this looks like it works, so I must be missing something. Is there a comment you could add to provide more info?

src/tape/ensemble.py
```diff
@@ -206,6 +209,54 @@ def info(self, verbose=True, memory_usage=True, **kwargs):
...
+        A boolean value indicating whether the sources tied to a given object
+        are only found in a single partition (True), or if they are split
+        across multiple partitions (False)
+
+        """
+        idx = ens._source.index
+        counts = idx.map_partitions(lambda a: Counter(a.unique())).compute()
+
+        unq_counter = counts[0]
+        for i in range(len(counts) - 1):
```
jeremykubica, line 254: might be clearer as `range(1, len(counts))` and then `unq_counter += counts[i]`

src/tape/ensemble.py
dougbrn: Thanks for calling this out, it's a vectorized operation that's actually doing element-wise comparison within the partition. As I was trying to put together an explanation, I realized that this is only doing comparison within the partition, and not across partitions. I've swapped this out to just use the existing `is_monotonic_increasing` function within the Dask/Pandas Index API, and expanded the test case to test the multi-partition case.

src/tape/ensemble.py
dougbrn: Added

PR #280, opened 2023-10-26
Fix EnsembleFrame.set_dirty and map_partitions metadata propagation

wilsonbb:
Previously `_Frame._propagate_metadata` called `result.set_dirty(self.is_dirty)` rather than `result.set_dirty(self.is_dirty())`.

This meant that when a transformation was performed on a dataframe, the result frame would return the bound *method* `is_dirty` itself rather than the boolean value returned by that method. In python, a boolean expression operating on a method itself interprets the method as `True`, resulting in unexpected cases where a transformed frame would incorrectly evaluate as dirty. This is fixed and some fields are renamed to avoid confusion.

In addition, `map_partitions` is overridden to propagate metadata if the output of the provided function is another` _Frame` object.


dougbrn (approved):
Nice find, this looks good to me

PR #281, opened 2023-11-01
Ensemble.update_frame no longer infers if a frame is dirty by checking if row count changed

wilsonbb:
Previously a call to`EnsembleFrame.update_frame(x)` for frame `x` would check if that frame was previously tracked, and if the row count had changed, would mark `x` as a dirty frame.

However, getting the row count through calls to `len` on a Dask dataframe can alter the task graph and cause lazy operations to be evaluated earlier. This was in particular causing significant slowdown in _sync_tables on a test notebook by forcing some lazy operations to be performed earlier. 

Here we move the calls to `set_dirty` to our methods wrapping Dask dataframe operations, including a new wrapper for `dropna`.

This may result in an increase in the number of syncs since even a `dropna` call on a table with no "na" values will now mark the table as dirty. However it should not result in more syncs than what is currently performed in the `main` branch.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -116,7 +116,7 @@ class Ensemble:
         self.update_frame(frame)
         return self
 
-    def update_frame(self, frame):
+    def update_frame(self, frame, check_dirty=False):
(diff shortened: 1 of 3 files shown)


dougbrn (approved):
Looks good to me

PR #283, opened 2023-11-04
Check Divisions, Enable Lazy Sync Operations on Divisions-enabled Ensembles, Overhaul Object Table Generation

dougbrn:
This PR does a few things that maybe I should have split up into different PRs, but there's also a decent amount of overlap. In this PR, I added more explicit support for Dask Divisions. In particular, checking if they are set after loading in data and warning the user if they are not. Divisions are also checked during sync operations to enable a new completely lazy sync that doesn't need to load the index fully into memory. In addition, the object table generation off the source table has been made lazy as well, removing the current unscalable implementation that loads the entire index into a pandas dataframe.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -678,8 +678,6 @@ class Ensemble:
 
         # Mask on object table
         self = self.query(f"{col_name} >= {threshold}", table="object")
-        # mask = self._object[col_name] >= threshold
-        # self._object = self._object[mask]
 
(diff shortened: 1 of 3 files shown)


wilsonbb (changes requested):
LGTM with a few suggestions

dougbrn:
@wilsonbb updated as per your suggestions, thank you!

wilsonbb (approved):

PR #285, opened 2023-11-06
Support storing batch results for custom meta

wilsonbb:
`Ensemble.batch` should track a result dataframe when the `label` parameter is used, and it should correctly translate the meta for the batch operation to be either an `TapeFrame` or `TapeSeries` (which Dask will use to map the batched result to either be an `EnsembleFrame` or `EnsembleSeries` respectively).

Note that `Ensemble.batch` still returns the result frame rather than the `Ensemble` to keep the API backwards compatible, but we can later have it return the `Ensemble` object to allow for easier chaining of commands.

Fixes a bug for creating meta for a non-empty EnsembleSeries.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1125,9 +1125,7 @@ class Ensemble:
                 # expensive computation.
                 warnings.warn(
                     f"""Warning: Batch successfully computed but failed to track frame under label,
-                      {label}, due to incompatible meta: {meta}""",
-                      label,
-                      meta)
+                      {label}, due to incompatible meta: {meta}""")
 
         if compute:
             return batch.compute()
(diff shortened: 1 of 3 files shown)


src/tape/ensemble.py
```diff
@@ -1021,6 +1021,9 @@ def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, **k
             Designates which column(s) to groupby. Columns may be from the
             source or object tables. For TAPE and `light-curve` functions
             this is populated automatically.
+        label: 'str', optional
+            If provided the ensemble will use this label to track the result 
+            dataframe. If `None`, the frame will not be tracked.
```
dougbrn, line 1026: Having the default be that the frame isn't tracked feels not ideal. Users may not set the name out of ignorance and then realize their result is not within the ensemble. Could we have this generate a dummy name by default "result1" that increments? And None disentangles it from the Ensemble as described above.

dougbrn (changes requested):
Looks good, just a comment on the best default for batch.

src/tape/ensemble.py
wilsonbb: Done.

dougbrn (approved):
Thanks! Looks good 😄

Issue #287, opened 2023-11-14
Query fails after #283 merge


nevencaplar:
This example fails after the latest merge 

```
from tape.ensemble import Ensemble
from tape.utils import ColumnMapper

datapath = "/astro/store/epyc3/data3/alerce_alerts/parquet/"

names = []
for i in range(12,36):
    name = datapath+f"source/lcs_00{i}.parquet"
    names.append(name)

ens = Ensemble()

colmap = ColumnMapper(id_col="oid",
                      time_col="mjd",
                      flux_col="magpsf_corr",
                      err_col="sigmapsf_corr_ext",
                      band_col="fid")
ens.from_parquet(source_file=names,
                #source_file=datapath+"source/lcs_000*.parquet",
                 #object_file=datapath+"object/*.parquet",
                 column_mapper=colmap,
                 sort=True)

ens.query("fid == 1", table="source").prune(30)
```
with `ValueError: This dataframe has npartitions=1, divisions should be a tuple of length=2, got 25`. The error does not happen if `sort=False`.

dougbrn:
Resolved by #288

PR #288, opened 2023-11-14
Calc nobs divisions

dougbrn:
Addresses #287. Makes calc_nobs use map partitions to propagate metadata instead of groupby when divisions are known. This introduces an annoying cost of having to calculate the unique bands for the resulting meta in the by_band=True case, it's slow, but works for now.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -615,7 +615,7 @@ class Ensemble:
 
                 # Get the band metadata
                 unq_bands = np.unique(self._source[band_col])
-                meta = {f"{band}": float for band in unq_bands}
+                meta = {band: float for band in unq_bands}
 
                 # Map the groupby to each partition
                 band_counts = self._source.map_partitions(
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -613,6 +613,10 @@ class Ensemble:
                 id_col = self._id_col
                 band_col = self._band_col
 
+                # Get the band metadata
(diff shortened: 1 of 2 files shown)


dougbrn:
pinging both @wilsonbb for general technical review and @nevencaplar to make sure this works for his use case.

src/tape/ensemble.py
```diff
@@ -607,41 +607,68 @@ def calc_nobs(self, by_band=False, label="nobs", temporary=True):
...
             # repartition the result to align with object
             if self._object.known_divisions:
-                band_counts.divisions = self._source.divisions
-                band_counts = band_counts.repartition(divisions=self._object.divisions)
+                # Grab these up front to help out the task graph
+                id_col = self._id_col
+                band_col = self._band_col
+
+                # Get the band metadata
+                unq_bands = np.unique(self._source[band_col])
```
wilsonbb, line 617: Nit: I'm not fully certain about the performance differences, but I feel `self._source[band_col].unique()` might be preferable and slightly more readable

wilsonbb (approved):
As discussed offline, it seems we should likely file an issue for adding a multi-partition data fixture in our testing suite. This hopefully would be more robust for catching issues like these.

Ultimately, looks good to me!

nevencaplar:
I have verified that this solves[ issue 287](https://github.com/lincc-frameworks/tape/issues/287)

nevencaplar (approved):

PR #295, opened 2023-11-22
Allow EnsembleFrame.compute to Trigger Object-Source Table Syncing

wilsonbb:
When `Ensemble.object.compute()` or `Ensemble.source.compute()` are called we would like to give the `Ensemble` a chance to perform an Object-Source table sync prior to computing the result.

Some previously tested workflows had other functions triggering syncs at the appropriate times (such as `Ensemble.batch`) or were using `Ensemble.compute()`. Now calling `EnsembleFrame.compute` should trigger any potential syncing for the tracking `Ensemble`.

Currently, we're preserving both the refactored `EnsembleFrame` API and the `Ensemble` API, so with this change I also update several unit tests to be parameterized to use both the refactored and 'legacy' methods (both `Ensemble.compute` and `EnsembleFrame.compute`, both`Ensemble.dropna` and `EnsembleFrame.dropna`, etc.) This is a bit awkward, and should be removed once the old API is deprecated.

### DIFF
--- a/tests/tape_tests/test_ensemble.py
+++ b/tests/tape_tests/test_ensemble.py
@@ -723,46 +723,84 @@ def test_update_column_map(dask_client):
     assert cmap_2.map["provenance_col"] == "p"
 
 
-def test_sync_tables(parquet_ensemble):
+@pytest.mark.parametrize("legacy", [True, False])
+def test_sync_tables(parquet_ensemble, legacy):
     """
-    Test that _sync_tables works as expected
+    Test that _sync_tables works as expected, using Ensemble-level APIs
+    when `legacy` is `True`, and EsnembleFrame APIs when `legacy` is `False`.
     """
-
-    assert len(parquet_ensemble.compute("object")) == 15
-    assert len(parquet_ensemble.compute("source")) == 2000
+    if legacy:
+        assert len(parquet_ensemble.compute("object")) == 15
+        assert len(parquet_ensemble.compute("source")) == 2000
+    else:
+        assert len(parquet_ensemble.object.compute()) == 15
+        assert len(parquet_ensemble.source.compute()) == 2000
 
     parquet_ensemble.prune(50, col_name="nobs_r").prune(50, col_name="nobs_g")
-    assert parquet_ensemble._object.is_dirty()  # Prune should set the object dirty flag
+    assert parquet_ensemble.object.is_dirty()  # Prune should set the object dirty flag
+
+    if legacy:
+        assert len(parquet_ensemble.compute("object")) == 5
+    else:
+        assert len(parquet_ensemble.object.compute()) == 5
 
     # Replace the maximum flux value with a NaN so that we will have a row to drop.
-    max_flux = max(parquet_ensemble._source[parquet_ensemble._flux_col])
-    parquet_ensemble._source[parquet_ensemble._flux_col] = parquet_ensemble._source[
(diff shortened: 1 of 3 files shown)


tests/tape_tests/test_ensemble.py
```diff
@@ -1130,8 +1250,10 @@ def test_select(dask_client):
     assert "count" not in ens._source.columns
     assert "something_else" not in ens._source.columns
 
-
-def test_assign(dask_client):
+@pytest.mark.parametrize("legacy", [True, False])
+def test_assign(dask_client, legacy):
+    """Tests assign for column-manipulation, using Ensemble.assign when `legacy` is `True`,
+    and EnsembleFrame.assign when `legacy` is `False`.
```
hombit, line 1256: Docstring is not closed

tests/tape_tests/test_ensemble.py
```diff
@@ -723,46 +723,84 @@ def test_update_column_map(dask_client):
...
+    when `legacy` is `True`, and EsnembleFrame APIs when `legacy` is `False`.
     """
-
-    assert len(parquet_ensemble.compute("object")) == 15
-    assert len(parquet_ensemble.compute("source")) == 2000
+    if legacy:
+        assert len(parquet_ensemble.compute("object")) == 15
+        assert len(parquet_ensemble.compute("source")) == 2000
+    else:
+        assert len(parquet_ensemble.object.compute()) == 15
```
hombit, line 736: What would happen if I do `parquet_ensemble.object.shape[0].compute()` instead? Do I understand correctly that our overloaded version of `compute()` will not be run?

hombit (changes requested):
Thank you, this has a perfect sense!

Please note, that GH actions didn't run for some reason. If they would run, they probably would detect a problem with syntax errors in `test_ensemble.py` and fail.

I have a general question: why do we care about object and source table synchronization and do not care about other frames?

I believe it adds a lot of implicitness. Let's say that I created a new "source" frame (it is "source" by semantics, not literarily `Ensemble.source`), accidentally or for purpose. How do I synchronies it to the object table?

tests/tape_tests/test_ensemble.py
wilsonbb: As discussed we will not be able to detect this here and there are other cases we would like to be able to handle as well.

Filed an issue for this in
https://github.com/lincc-frameworks/tape/issues/296

tests/tape_tests/test_ensemble.py
wilsonbb: Fixed.

wilsonbb:
> Thank you, this has a perfect sense!
> 
> Please note, that GH actions didn't run for some reason. If they would run, they probably would detect a problem with syntax errors in `test_ensemble.py` and fail.
> 
> I have a general question: why do we care about object and source table synchronization and do not care about other frames?
> 
> I believe it adds a lot of implicitness. Let's say that I created a new "source" frame (it is "source" by semantics, not literarily `Ensemble.source`), accidentally or for purpose. How do I synchronies it to the object table?

The short version, is that this was part of the initial design, with an `Ensemble.link_frames(frame1, frame2)` method. This would allow any arbitrary frame to be synced the way Object and Source are. Since there isn't immediate demand this is unimplemented, but have filed https://github.com/lincc-frameworks/tape/issues/297 to track that feature.

hombit (approved):

PR #301, opened 2023-11-29
Add Explicit Metadata Propagation for EnsembleFrame joins 

wilsonbb:
Overrides [`dask.dataframe.DataFrame.join`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.join.html) so that we explicitly propagate metadata.

This fixes an error that happens when syncing tables with `known_divisions` introduced by merging PR https://github.com/lincc-frameworks/tape/pull/288 into this refactor branch from main.

Also note that `frame1.join(frame2)` will yield a result of type `frame1` regardless of the value of the `how` parameter and explicit testing of this is added here.

### DIFF
--- a/src/tape/ensemble_frame.py
+++ b/src/tape/ensemble_frame.py
@@ -286,6 +286,12 @@ class _Frame(dd.core._Frame):
         result = super().merge(right, **kwargs)
         return self._propagate_metadata(result)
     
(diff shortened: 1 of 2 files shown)


dougbrn (approved):
This looks good to me, I think always having frame1.join(frame2) return a `frame1`-like object seems correct.

Issue #137, opened 2023-06-15
Add method .__version__


nevencaplar:
Add method to show version so that code:

`import tape; tape.__version__`

creates a valid output.

dougbrn:
Closed by #302 , @nevencaplar just fyi.

PR #302, opened 2023-11-29
Add version

dougbrn:
Addresses #137, looks like we were just missing the one line in the init, as setuptools already dynamically generates a _version.py file.

(Should be completely non-conflicting with the refactor)


wilsonbb (approved):
A small but very helpful change!

PR #303, opened 2023-11-30
Fix table syncing to use inner joins.

wilsonbb:
In https://github.com/lincc-frameworks/tape/pull/283, we introduced logic to allow for lazy syncing of tables with known divisions. This was accomplished via a [`join`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.join.html) call which was a left join by default.

However since we always sync the object table before the source table, it's possible to create an arrangement of operations values removed from the source table's index can be synced back from a dirty object table. This PR updates `test_ensemble.test_sync_tables` to have such a failing case and fixes it by switching to inner joins.


dougbrn (approved):
Looks good, thanks for this!

PR #322, opened 2023-12-11
remove provenance machinery for now

### LINKED ISSUE #292: Remove provenance machinery
Provenance was added a while ago as a first step towards the multi-survey problem. However, that's further down the road for now and provenance is introducing some API bloat and slowdown. I'd advocate for scrapping it and re-implementing it fresh when there is more purpose to it.

dougbrn:
Resolves #292


wilsonbb (approved):
LGTM!

Issue #317, opened 2023-12-09
Improve Per-Band Support in `Ensemble.batch`


wilsonbb:
It's pretty typical for a user to want to perform per-band analysis via `Ensemble.batch`. Currently, a user can implement themselves in their custom function via array masking, but this is a) poor user experience b) gives the user opportunity to make an error.

Currently, `Ensemble.batch`'s `on` parameter could be used to allow the user to specify that the results should be grouped by both object ID and band. This yields an output table with a `MultiIndex` which (in my attempts at playing around with TAPE) initially works but eventually produces an error in the computation graph (unsurprising since Dask lacks [MultiIndex support](https://github.com/dask/dask/issues/1493)).

Possible solutions include:
* @dougbrn has suggested that we could consider using a pivot table in `Ensemble.batch` similar to what we do in `Ensemble.calc_nobs`
* Provide a specific `per_band` boolean parameter to `Ensemble.batch` that will take a user's output columns and create one per-each observed band (though we probably want to do some variation of the above anyway).

hombit:
I like this idea, it would simplify feature extraction support. However, I see some discussion points here. Would we like user to select a list of bands? Do we really want it in `batch`?

1. I don't really understand the first proposal...
2. It would make `batch` to be quiet complicated, especially in `meta` specification by user and return types. I would think about alternatives, for example having a separate `batch_per_band` or something like that.

PR #325, opened 2023-12-13
Allow to ignore band with FeatureExtractor

### LINKED ISSUE #323: Feature extraction should be able to ignore passband
Currently, built-in analysis feature extraction function requires mandatory `band_to_calc` kwarg to specify which passband to consider. It doesn't help when user wants to ignore band and apply function to the whole light curve.

We should support `band_to_calc=None` which would ignore passband.

hombit:
Currently, `FeatureExtractor` analysis function requires `band_to_calc` parameter to specify passband it would pre-filter light curve on. This PR introduces `band_to_calc=None`, which tells the function to ignore passband information and apply the function to the whole light curve.

This option could be useful to some types of analysis, for example `light_curve.OtsuSplit` is not so sensitive to passband if it is applied to differential fluxes.

Another use case is datasets where each object is observed in a single passband, for example ZTF DRs. In such cases it could be practically useful to ignore passband information.

This PR is related to #317, but covers feature extraction only. `FeatureExtractor` could should be revisited again if/when #317 is implemented.

Fixes #323

### DIFF
--- a/docs/tutorials/working_with_the_ensemble.ipynb
+++ b/docs/tutorials/working_with_the_ensemble.ipynb
@@ -750,6 +750,7 @@
     "import light_curve as licu\n",
     "\n",
     "extractor = licu.Extractor(licu.Amplitude(), licu.AndersonDarlingNormal(), licu.StetsonK())\n",
+    "# band_to_calc=None will ignore the band column and use all sources for each object\n",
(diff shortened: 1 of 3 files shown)


dougbrn (approved):
looks good!

PR #327, opened 2023-12-14
Add `by_band` kwarg to batch, allow each lightcurve band to be processed independently

### LINKED ISSUE #241: Ensemble.batch should not compute by default
Computing by default is not good behavior, as it limits the ability to work with the result alongside the Ensemble (column assignment to object/source for example). It's also just very against the Dask mantra of lazy evaluation. We should at least make this default to False, but I could even see just removing the kwarg and requiring the user to specifically call compute on the result if they want an in-memory result.

### LINKED ISSUE #317: Improve Per-Band Support in `Ensemble.batch`

dougbrn:
Resolves #317
Resolves #241 

A bit of scope creep on this PR, the main purpose is to address #317, giving more support to users looking to apply a function to each band within a lightcurve within the batch interface. This is done by adding the `by_band` kwarg to `Ensemble.batch`, which makes sure that the band column is part of the groupby and also modifies the output to return an output column for each band. As a result, this also standardizes the output of batch to always return an EnsembleFrame object (even when this keyword is not used), where as previously in some cases it would return a series.

I decided to go ahead with addressing #241 as part of this as well, removing the `compute` kwarg within `Ensemble.batch`, I opted not to replace it with persist, as I believe that batch should just return a lazy object that users can chain into a compute or persist command as they see fit.

I also went ahead and added a batch showcase tutorial as part of this PR, in that I tried to outline how this change actually is used, and also tried to be transparent about some issues regarding dataframe outputs.


src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
                 meta=meta,
             )
 
+        # Output standardization
```
hombit, line 1163: It would be helpful for readability if you could brake this code into few functions

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+        if isinstance(batch, EnsembleSeries):
+            if batch.name == self._id_col:
+                batch = batch.rename("result")
+            batch = EnsembleFrame.from_dask_dataframe(batch.to_frame())
+            if len(on) > 1:
+                batch = batch.reset_index()
+                # Need to overwrite the meta manually as the multiindex will be
+                # interpretted by dask as a single "index" column
+                batch._meta = TapeFrame(columns=on + ["result"])
```
hombit, line 1172: 1. Do we need dtypes here?
2. What if previously, on line 1165 `batch.name != self._id_col`? Would we have "result" column in that case?

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+
+                if by_band:
+                    batch = batch.categorize(self._band_col).pivot_table(
+                        index=on[0], columns=self._band_col, aggfunc="sum"
+                    )
```
hombit, line 1177: Could you please explain this part? Why do we use `aggfunc="sum"` here? Why could/should we use `index=on[0]` here?

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+
+                    # Need to once again reestablish meta for the pivot
+                    band_labels = batch.columns.values
+                    out_cols = []
+                    for col in ["result"]:
```
hombit, line 1182: Why do we need this loop here?

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+                        for band in band_labels:
+                            out_cols += [(str(col), str(band))]
+                    batch._meta = TapeFrame(columns=out_cols)
+                    # Flatten the columns to a new column per band
+                    batch.columns = ["_".join(col).strip() for col in batch.columns.values]
```
hombit, line 1187: Why do we need this `str.strip()` here?

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
                 meta=meta,
             )
 
+        # Output standardization
+        if isinstance(batch, EnsembleSeries):
+            if batch.name == self._id_col:
+                batch = batch.rename("result")
+            batch = EnsembleFrame.from_dask_dataframe(batch.to_frame())
+            if len(on) > 1:
+                batch = batch.reset_index()
```
hombit, line 1169: Should we here and further do it in-place? I guess that it wouldn't change performance for Dask Dataframes (while it could help with pandas), but I'm not sure...

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+
+                    # Flatten the columns to a new column per band
+                    batch.columns = ["_".join(col).strip() for col in batch.columns.values]
+
+                    # The pivot returns a dask dataframe, need to convert back
+                    batch = EnsembleFrame.from_dask_dataframe(batch)
+
+                else:
+                    batch = batch.set_index(on[0], sort=False)
+
```
hombit, line 1221: Do we need to check some third case after `isinstance(batch, EnsembleSeries)` and `isinstance(batch, EnsembleFrame)`?

src/tape/ensemble.py
```diff
@@ -1142,6 +1160,65 @@ def s2n_inter_quartile_range(flux, err):
...
+                batch._meta = TapeFrame(columns=on + res_cols)
+                if by_band:
+                    batch = batch.categorize(self._band_col)
+                    batch = batch.pivot_table(index=on[0], columns=self._band_col, aggfunc="sum")
+
+                    # Need to once again reestablish meta for the pivot
+                    band_labels = batch.columns.values
+
+                    out_cols = []
+                    for col in res_cols[::-1]:
```
hombit, line 1208: Why do we do it in a reverse way?

hombit (commented):
Thank you!

I have few clarification questions and a request to make `batch()` more readable with helper functions

src/tape/ensemble.py
dougbrn: This is a consequence of only having access to `pivot_table` from pandas when it would be ideal to have `pivot` or `unstack`. With the way that batch sets up it's grouping, I don't believe the aggregation here will ever do any work.

Setting the index on "on[0]" lets pivot_table know that the index of the result should be the first grouping key, which seemed like the ideal behavior as the first key is used as main index in the non by-band case.

hombit: Thank you!

src/tape/ensemble.py
dougbrn: good catch, I think I left this open to further generalization but for the EnsembleSeries case it should always just be a single result, will remove this.

src/tape/ensemble.py
dougbrn: don't think it's a need, but seemed good to not preserve any errant whitespace

src/tape/ensemble.py
dougbrn: Those should encompass the only real outcomes, but maybe good to add an else that raises an error just for completeness.

src/tape/ensemble.py
```diff
@@ -1110,6 +1118,16 @@ def s2n_inter_quartile_range(flux, err):
         if isinstance(on, str):
             on = [on]  # Convert to list if only one column is passed
 
+        if by_band:
+            if self._band_col not in on:
+                on += [self._band_col]
+            elif (
+                on[-1] != self._band_col
+            ):  # Unsure if this is neccesary but it seems likely to produce wrong behavior
```
wilsonbb, line 1126: Can we always ensure correct `on` ordering for the user? We're doing a linear search of `on` anyway and this would prevent an error and let us remove a weird caveat from the docstring.

We also 

I think the code could look something like

```
if by_band:
    if self._band_col not in on:
        on += [self._band_col]
    elif on[-1] != self._band_col:
        # Ensure band is the final column in the `on` list
        on[on.index(self._band_col)] = on[-1]
        on[-1] = self._band_col
```

Alternatively could use `list.index()` to get the index and check if present in `on` in one pass, but would need a try-except for handling if it's missing from the list (which would replace the if-elif here)

docs/tutorials/batch_showcase.ipynb
```diff
@@ -0,0 +1,578 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def my_bounds_df(flux):\n",
+    "    return pd.DataFrame({'min':[np.min(flux)], 'max':[np.max(flux)]})\n",
```
wilsonbb, line 298: I might be confused here but am wondering if returning 

```
    return pd.Series([np.min(flux), np.max(flux)], index=['min', 'max'])
```

resolves some of the weird re-indexing issue. It's a bit counter-intuitive, but this is what's recommended in the [`pd.DataFrame.apply` docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html)

src/tape/ensemble.py
dougbrn: Just because this is the way that Pandas appears to do it, and the ordering of the columns matters for meta I believe.

src/tape/ensemble.py
dougbrn: 1. I'm not 100% sure, Dask's dtype inference seems much more robust than it's column label inference, so I haven't encountered any issues with leaving dtypes out. But it's possible there are some failure states I'm not accounting for.
2. We would not, and that's a good point as it might cause problems downstream in that case since we reference "result", maybe switching from result to batch.name throughout is more appropriate here

docs/tutorials/batch_showcase.ipynb
dougbrn: I'm assuming you mean to return a dataframe here? Returning a series with the index set does resolve this, but that's basically what case 2 shows. For users who have some pre-established functionality that returns a dataframe, the issue seems to persist even if you try to set the index as in that example (see attached). At least, that's what I'm trying to show with case 3

Screen Shot 2023-12-19 at 2 50 55 PM

src/tape/ensemble.py
dougbrn: This sounds reasonable, will try it out!

src/tape/ensemble.py
dougbrn: I think the answer is no. For one it's not supported in Dask, but also it even really isn't advisable in Pandas it seems... https://stackoverflow.com/questions/43893457/understanding-inplace-true-in-pandas

docs/tutorials/batch_showcase.ipynb
wilsonbb: Ah that makes sense to me!

src/tape/ensemble.py
dougbrn: I've broken all the formatting into a separate function, let me know if you think this should go further.

src/tape/ensemble.py
dougbrn: Addressed 2. in latest commit

src/tape/ensemble.py
dougbrn: resolved by latest commit

src/tape/ensemble.py
dougbrn: addressed in latest commit

src/tape/ensemble.py
dougbrn: Implemented in latest commit

src/tape/ensemble.py
hombit: Thank you

src/tape/ensemble.py
hombit: Thank you, `batch` is much more readable now! That new function is still a bit hard to understand, may you please add more comments/helper functions there?

src/tape/ensemble.py
hombit: Thank you!

src/tape/ensemble.py
hombit: I don't know what is the right principle here: format new column names in a predictable way (if users access them programmatically) or make columns to have nice names (if users would type them).

We never trim column names in other places, maybe it would be better to have more predictable behavior here?

src/tape/ensemble.py
```diff
@@ -1032,20 +1032,26 @@ def batch(self, func, *args, meta=None, use_map=True, compute=True, on=None, lab
             the results. Overridden by TAPE for TAPE and `light-curve`
             functions. If none, attempts to coerce the result to a
             pandas.Series.
+        by_band: `boolean`
```
hombit, line 1035: Could you please add here a sentence about output column names?

src/tape/ensemble.py
hombit: Thank you, sometimes Pandas' interfaces has so little sense...

docs/tutorials/batch_showcase.ipynb
```diff
@@ -0,0 +1,578 @@
+{
```
hombit, line 1: <p>I believe that we should prefer the following pattern for array building, because <code>np.append</code> allocates too much. I understand that practically in this case difference is small, however I would prefer to give best-practice code in the docs.</p><p> </p><pre class="ql-syntax" spellcheck="false">%%timeit
obj_ids = []
for i in range(10,110):
    obj_ids.append(np.array([i]*1250))
obj_ids = np.concatenate(obj_ids)
  
3.76 ms ± 18.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
</pre><p>Compare to</p><pre class="ql-syntax" spellcheck="false">%%timeit
obj_ids = np.array([])
for i in range(10,110):
    obj_ids = np.append(obj_ids,np.array([i]*1250))
5.66 ms ± 114 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)  
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lincc-frameworks/tape/pull/327/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

src/tape/ensemble.py
dougbrn: I'm happy to remove the strip call, we can always add it later if there are clear use cases for it

src/tape/ensemble.py
dougbrn: Will do, but just to try to illustrate for your understanding, in this screenshot you basically see the effect of a before and after of pandas.pivot_table. Notice that before we had the columns as (min, max), but after we have (max, min) as the main ordering structure of the new columns. All the reverse is doing is making sure the meta columns reflect this.

Screen Shot 2023-12-20 at 11 49 33 AM

dougbrn: added comment in latest commit

src/tape/ensemble.py
dougbrn: I rewrote the sentence that tried to explain how the columns affected, let me know if that  adds any clarity

docs/tutorials/batch_showcase.ipynb
dougbrn: implemented

src/tape/ensemble.py
dougbrn: I've paired down the logic and added additional comments in the latest commit. Hopefully that helps!

src/tape/ensemble.py
dougbrn: removed in latest commit

src/tape/ensemble.py
wilsonbb: Thank you!

wilsonbb (approved):
LGTM. Thanks for doing this, Doug!

hombit (approved):
Look great, thank you so much!

PR #336, opened 2023-12-21
Adds a random lightcurve selection function to the Ensemble

### LINKED ISSUE #332: Add a random lightcurve selection function
We should have an Ensemble function to select a random lightcurve to inspect. This seems like a nice feature for validation/sanity check that a workflow has returned the expected result.

My thinking is to just select a random object ID from one partition (at certain scales loading the whole index may be problematic, randomly select a single partition to search), and then `loc` that in source.

This should return a timeseries object

dougbrn:
Resolves #332.

Adds a convenience function to grab a random lightcurve from the Ensemble, which we expect users to use to do quick sanity checks that operations have had the desired effect on their lightcurves.

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1870,10 +1870,8 @@ class Ensemble:
         in larger partitions.
 
         """
-        if seed is not None:
-            rng = np.random.default_rng(seed)
-        else:
-            rng = np.random
+
+        rng = np.random.default_rng(seed)
 
         # We will select one partition at random to select an object from
         partitions = np.array(range(self.object.npartitions))
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1848,7 +1848,7 @@ class Ensemble:
         return self
 
     def select_random_timeseries(self, seed=None):
-        """Selects a random lightcurve from the Ensemble
+        """Selects a random lightcurve from a random partition of the Ensemble.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


src/tape/ensemble.py
```diff
@@ -1847,6 +1847,36 @@ def _sync_tables(self):
...
+        -------
+        ts: `TimeSeries`
+            Timeseries for a single object
+
+        """
+
+        if seed is not None:
+            np.random.seed(seed)
+
+        # Avoid a choice from full index space, select a random partition to grab from
```
hombit, line 1869: Do we know number of objects per partition here? It would be nice to select an integer randomly over all objects, and then get a partition where our integer index is located.

src/tape/ensemble.py
```diff
@@ -1847,6 +1847,36 @@ def _sync_tables(self):
...
+        if self.object.npartitions > 1:
+            partition_num = np.random.randint(0, self.object.npartitions - 1)
+            partition_ids = self.object.get_partition(partition_num).index.values
```
hombit, line 1872: What if partition is empty?

src/tape/ensemble.py
```diff
@@ -1847,6 +1847,36 @@ def _sync_tables(self):
...
+            lcid = np.random.choice(partition_ids)
+        else:
+            partition_num = 0
+            lcid = np.random.choice(self.object.index.values)
+        print(f"Selected Object {lcid} from Partition {partition_num}")
```
hombit, line 1877: Do we need this `print` here?

hombit (changes requested):

src/tape/ensemble.py
dougbrn: Wilson brought this up too, it's a good point which I'll need to consider more, might need to have it try and choose again if it's empty.

src/tape/ensemble.py
dougbrn: We don't and that will be expensive at scale (to grab lengths)

src/tape/ensemble.py
dougbrn: I thought it would be good to convey to the users what was grabbed, as that information isn't presented front-and-center in the resulting timeseries object

src/tape/ensemble.py
```diff
@@ -1847,6 +1847,36 @@ def _sync_tables(self):
...
+
+        Returns
+        -------
+        ts: `TimeSeries`
+            Timeseries for a single object
+
+        """
+
+        if seed is not None:
+            np.random.seed(seed)
```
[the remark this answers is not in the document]
dougbrn: sounds good, will do

src/tape/ensemble.py
hombit: Sounds reasonable! Could you please update the docstring to mention that we do **not** sample uniformly here?

src/tape/ensemble.py
```diff
@@ -1847,6 +1847,36 @@ def _sync_tables(self):
...
+
+        # Avoid a choice from full index space, select a random partition to grab from
+        if self.object.npartitions > 1:
+            partition_num = np.random.randint(0, self.object.npartitions - 1)
```
hombit, line 1871: We never select the last partition here, because `np.random.randint`'s second argument is an exclusive high value. Actually, I'm not sure we even need `if-else` here: for `self.object.npartitions == 1` `randint(0, 1)` will always return zero.

src/tape/ensemble.py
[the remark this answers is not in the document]
dougbrn: Done

src/tape/ensemble.py
dougbrn: Done

src/tape/ensemble.py
dougbrn: In the latest commit, I took @wilsonbb's suggestion to first shuffle the partitions and then scan through, checking the index lengths until finding a non-empty partition. Added a unit test to cover this case.

src/tape/ensemble.py
dougbrn: Should be addressed by the refactor done in the latest commit

src/tape/ensemble.py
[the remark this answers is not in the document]
hombit: For the sake of the simplification you could also drop if-else here: `np.random.rng(None)` creates a new generator with a random seed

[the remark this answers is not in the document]
dougbrn: Good to know, will do!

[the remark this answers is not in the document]
dougbrn: Done

hombit (approved):
Thank you!

PR #340, opened 2024-01-02
Update to Python Project Template v1.4.5; Integrate with ASV Benchmarking

dougbrn:
The vast majority of lines added/removed here are from the copier update command (grabbing the latest and greatest from ppt 1.4.5). This PR adds a few starter benchmarks for TAPE ASV integration. The github-actions comment below demonstrates that the PR comparison workflow is working as expected.

However, I don't think there's a way to test the gh-pages deployment until this gets merged into main, based on that I would like to try to merge this with a minimal benchmarking suite, verify whether the page deployment works as expected, and then build out the benchmarking suite in a later PR.


.github/pull_request_template.md
drewoldag, line 1: Don't feel obligated to keep this pull request template unless you think it will be useful for the project. Some people have expressed concern about how wordy it is.

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,42 @@
+"""Two sample benchmarks to compute runtime and memory usage.
```
drewoldag, line 1: Low priority, update the docstring here.

drewoldag (approved):
This looks ok to me. Neither of the comments that I left should be considered blockers.

.github/pull_request_template.md
dougbrn: We can try it out for a bit! I removed a few lines that seemed redundant from CI results (e.g. my code passes tests is self-evident by the PR)

PR #343, opened 2024-01-08
Save to parquet

### LINKED ISSUE #151: Save Ensemble to Parquet
Saving the ensemble to parquet would be useful for saving state. The object and source tables should probably be saved as two separate parquet files.

Neccessary Components:

- [x] ensemble.save_ensemble() function; Saves an Ensemble to disk, with subdirectories for object and source partitions. Option (perhaps on by default) to include all result frames, or just a subset of result frames
- [x] ensemble.from_ensemble() function; Reads the directory structure established by save (parquet-like) and loads into a new ensemble, I think this will need to be a separate function from read_parquet to handle the addition of many result tables implicitly
- [x] from_ensemble(); An Ensemble constructor function that bypasses ensemble initialization
- [x] unit tests; with some consideration to how we test the save function within the actions framework. Does it just purge any created files automatically? _Edit: No it doesn't. Resolved this by using pytest with pathlib.Path temporary directories_
- [x] documentation; this feature will be important enough to need some documentation beyond just autoapi.

Non-critical Components:
- [x] Potentially save the column mapper and load it within the save/load internals. Avoid needing to specify the column mapper in the load.

Open Questions/Issues:
* Schema error with saving Stetson_J result frames to parquet. These frames have a dictionary result column, it may be that there's a simple schema argument to pass, but this would also be resolved by just making stetson_J return a column per band (as has been talked about for a while). Given that stetson_J is really just a tester function for us, I'm opting to not spend time on getting it working.

dougbrn:
## Change Description
Resolves #151.

## Code Quality

## Project-Specific Pull Request Checklists
- [x] I have added a function that requires a sync_tables command, and have added the neccesary sync_tables call

### Bug Fix Checklist
- [ ] My fix includes a new test that breaks as a result of the bug (if possible)
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### New Feature Checklist
- [x] I have updated the tutorial to highlight my new feature (if appropriate)
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### Documentation Change Checklist
- [x] Any updated docstrings use the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html)

### Build/CI Change Checklist
- [ ] If required or optional dependencies have changed (including version numbers), I have updated the README to reflect this
- [ ] If this is a new CI setup, I have added the associated badge to the README

### Other Change Checklist
- [x] Any new or updated docstrings use the [NumPy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] I have updated the tutorial to highlight my new feature (if appropriate)
- [ ] I have added unit/End-to-End (E2E) test cases to cover any changes
- [ ] My change includes a breaking change
  - [ ] My change includes backwards compatibility and deprecation warnings (if possible)

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1297,6 +1297,13 @@ class Ensemble:
             # grab the dataframe from the frame label
             frame = self.frames[frame_label]
 
+            # Object can have no columns, which parquet doesn't handle
+            # In this case, we'll avoid saving to parquet
+            if frame_label == "object":
+                if len(frame.columns) == 0:
+                    print("The Object Frame was not saved as no columns were present.")
+                    continue
+
             # creates a subdirectory for the frame partition files
             frame.to_parquet(os.path.join(ens_path, frame_label), **kwargs)
 
@@ -1342,8 +1349,13 @@ class Ensemble:
 
         # Load Object and Source
         obj_path = os.path.join(dirpath, "object")
-        src_path = os.path.join(dirpath, "object")
-        self.from_parquet(src_path, obj_path, column_mapper=column_mapper, **kwargs)
+        src_path = os.path.join(dirpath, "source")
+
+        # Check for whether or not object is present, it's not saved when no columns are present
+        if "object" in os.listdir(dirpath):
+            self.from_parquet(src_path, obj_path, column_mapper=column_mapper, **kwargs)
+        else:
+            self.from_parquet(src_path, column_mapper=column_mapper, **kwargs)
 
         # Load all remaining frames
         if additional_frames is False:
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1301,7 +1301,7 @@ class Ensemble:
 
         return
 
-    def from_ensemble(self, dirpath, additional_frames):
+    def from_ensemble(self, dirpath, additional_frames=True, column_mapper=None, **kwargs):
         """Load an ensemble from an on-disk ensemble.
 
         Parameters
@@ -1316,13 +1316,47 @@ class Ensemble:
             should be loaded. Object and Source will always be added and do not
             need to be specified in the list. By default, all frames will be
             loaded.
+        column_mapper: Tape.ColumnMapper object, or None, optional
+            Supplies a ColumnMapper to the Ensemble, if None (default) searches
+            for a column_mapper.npy file in the directory, which should be
+            created when the ensemble is saved.
 
         Returns
         ----------
         ensemble: `tape.ensemble.Ensemble`
             The ensemble object.
         """
-        pass
+
+        # First grab the column_mapper if not specified
+        if column_mapper is None:
+            column_mapper = np.load(os.path.join(dirpath, 'column_mapper.npy'), allow_pickle='TRUE').item()
+
+        # Load Object and Source
+        obj_path = os.path.join(dirpath, "object")
+        src_path = os.path.join(dirpath, "object")
+        self.from_parquet(src_path, obj_path, column_mapper=column_mapper, **kwargs)
+
+        # Load all remaining frames
+        if additional_frames is False:
+            return self  # we are all done
+        else:
+            if additional_frames is True:
+                #  Grab all subdirectory paths in the top-level folder, filter out any files
+                frames_to_load = [f for f in os.listdir(dirpath) if not os.path.isfile(os.join(dirpath, f))]
+            elif isinstance(additional_frames, Iterable):
+                frames_to_load = [os.path.join(dirpath, frame) for frame in additional_frames]
+            else:
+                raise ValueError("Invalid input to `additional_frames`, must be boolean or list-like")
+
+            # Filter out object and source from additional frames
+            frames_to_load = [frame for frame in frames_to_load if os.path.split(frame)[1] != "object" or os.path.split(frame)[1] != "source"]
+
+            for frame in frames_to_load:
+                label = os.path.split(frame)[1]
+                ddf = EnsembleFrame.from_parquet(frame, label=label)
+                self.add_frame(ddf, label)
+
+            return self
 
     def from_pandas(
         self,
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -8,6 +8,7 @@ import pandas as pd
 
 from dask.distributed import Client
 from collections import Counter
+from collections.abc import Iterable
 
 from .analysis.base import AnalysisFunction
 from .analysis.feature_extractor import BaseLightCurveFeature, FeatureExtractor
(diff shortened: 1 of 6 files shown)


src/tape/ensemble.py
```diff
@@ -1237,6 +1240,165 @@ def _standardize_batch(self, batch, on, by_band):
...
+
+        Note
+        ----
+        If the object frame has no columns, which is often the case when an
+        Ensemble is constructed using only source files/dictionaries, then an
+        object subdirectory will not be created. `Ensemble.from_ensemble` will
+        know how to work with the directory whether or not the object
+        subdirectory is present.
+
+        Be careful about repeated saves to the same directory name. This will
```
wilsonbb, line 1275: Given how liberally jupyter users may hit "run all cells", I think this could lead to some easy mistakes.

What about having an "overwrite" parameter?

If True, we will first try to delete the save directory (possibly printing a message if we did so)

If False, we can throw an exception if the save directory already exists noting they should use a different name or set "overwrite=True"

My instinct is that the default value of the parameter should be False.

src/tape/ensemble.py
```diff
@@ -1237,6 +1240,165 @@ def _standardize_batch(self, batch, on, by_band):
...
+            column_mapper = ColumnMapper()
+            column_mapper.map = map_dict
+
+        # Load Object and Source
+        obj_path = os.path.join(dirpath, "object")
+        src_path = os.path.join(dirpath, "source")
+
+        # Check for whether or not object is present, it's not saved when no columns are present
+        if "object" in os.listdir(dirpath):
+            self.from_parquet(src_path, obj_path, column_mapper=column_mapper, **kwargs)
```
wilsonbb, line 1370: Should we always use `sync_tables=False` since we synced object and source before saving?

Also as a general question that may be out of scope for this PR, are we preserving divisions here? If so I guess it would be helpful to use `sorted=True` if we knew that information?

src/tape/ensemble.py
dougbrn: Yeah, so we have access to overwrite already via the **kwargs passed along to all the to_parquet calls. What remains is that that will only overwrite any subdirectories (ensemble/object, ensemble/source, ensemble/result, etc.) that were present in both saves. So any previous subdirectories will not be deleted if a new save doesn't include them. I'm really worried about doing any form of delete, just as someone could specify a non-empty directory and I wouldn't want to delete anything else. 

So for 1. Maybe the overwrite parameter should graduate out of the **kwargs for visibility? And 2. I'm not sure but maybe we could have save write some kind of subdirs metadata file that lets TAPE know which subdirectories it made, so that overwrite can clean those directories up in subsequent overwrite-enabled saves?

src/tape/ensemble.py
dougbrn: I believe we can always have `sync_tables=False`, right, good point!

For divisions, these are not being preserved at the moment, which definitely seems like an issue. We can save an ensemble without divisions being known, so the sorted parameter will need to be dependent on that information. 
Similar to above, maybe there's a need for a metadata file that should be generated to indicate whether divisions are known that can be read in?

dougbrn: I just pushed another commit to better expose the sort/sorted kwargs from ensemble.from_parquet so they are usable in these cases. Some kind of metadata may be preferable here still so the user doesn't need to specify at all.

src/tape/ensemble.py
dougbrn: Wilson and I talked offline a bit on this. The strategy that we agreed on is to generate a metadata file that knows which subdirectories were produced by the save command. Successive save commands will look for this file and use it to locate which subdirectories in the directory should be removed during the save operation. In particular, we'd be looking for any subdirectories created by a previous save command that will not be overwritten by the current save command. Parquet will handle overwrites for conflicting directories, but we will need ensemble.save_ensemble to catch and remove these frame subdirectories that would otherwise not be touched.

dougbrn: The latest commit implements this, with the tweak that I ended up opting to just clean all the previous subdirectories up first as it was logically easier.

src/tape/ensemble.py
```diff
@@ -1237,6 +1240,165 @@ def _standardize_batch(self, batch, on, by_band):
...
+        # Determine the path
+        ens_path = os.path.join(path, dirname)
+
+        # Compile frame list
+        if additional_frames is True:
+            frames_to_save = list(self.frames.keys())  # save all frames
+        elif additional_frames is False:
+            frames_to_save = ["object", "source"]  # save just object and source
+        elif isinstance(additional_frames, Iterable):
+            frames_to_save = [frame for frame in additional_frames if frame in list(self.frames.keys())]
```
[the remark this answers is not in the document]
dougbrn: Implemented in latest commit

src/tape/ensemble.py
dougbrn: Another update, I now have this working with metadata. ensemble.save_ensemble now stores a boolean per frame in the metadata.json file indicating whether that frame had divisions set. Instead of using the sort/sorted kwargs, I've changed ensemble.from_ensemble to use this to determine whether to calculate divisions or not. In the case of object and source, this is just by setting the sorted flag or not. For the other frames, the sorted flag is only applicable to the set_index call that ensemble.from_parquet folds into. So this PR now has the frames generate a parquet metadata file which populates divisions information if it's available, and the reading of additional frames now uses the parquet metadata file. 

I experimented with trying to get our ensemble.from_parquet to use these _metadata files, but it would be a non-trivial change in the logic that felt out of scope for this PR. It's potentially worth logging for the future as using the _metadata files would avoid having to calculate the min/max values per partition to get divisions, but on the other hand these files can get quite large for large datasets so they may have issues at scale.

src/tape/ensemble.py
```diff
@@ -1499,6 +1559,12 @@ def from_dask_dataframe(
         self._load_column_mapper(column_mapper, **kwargs)
         source_frame = SourceFrame.from_dask_dataframe(source_frame, self)
 
+        # Repartition before any sorting
+        if npartitions and npartitions > 1:
+            source_frame = source_frame.repartition(npartitions=npartitions)
+        elif partition_size:
+            source_frame = source_frame.repartition(partition_size=partition_size)
```
wilsonbb, line 1566: I think we can just add a partition_size pytest parameter and we can get rid of the code coverage warning here

src/tape/ensemble_readers.py
```diff
@@ -10,7 +10,15 @@
 from tape.utils import ColumnMapper
 
 
-def read_ensemble(dirpath, additional_frames=True, column_mapper=None, dask_client=True, **kwargs):
+def read_ensemble(
+    dirpath,
+    additional_frames=True,
+    column_mapper=None,
+    dask_client=True,
```
wilsonbb, line 17: Nit: move dask_client to reflect Parameter order in the doc string (or vice versa)

src/tape/ensemble_readers.py
```diff
@@ -29,6 +37,19 @@ def read_ensemble(dirpath, additional_frames=True, column_mapper=None, dask_clie
...
+        containing the critical quantities (id,time,flux,err,band)
+    partition_size: `int`, optional
+            If specified, attempts to repartition the ensemble to partitions
+            of size `partition_size`.
+    sorted: bool, optional
+        If the index column is already sorted in increasing order.
+        Defaults to False
+    sort: `bool`, optional
+        If True, sorts the DataFrame by the id column. Otherwise set the
+        index on the individual existing partitions. Defaults to False.
```
wilsonbb, line 52: Do we want `sort` and `sorted` to be explicit parameters in the function definition? Or did we mean to remove these from the docstring?

wilsonbb (approved):
Have some questions and suggestions but overall looks good to me! Thanks Doug!

src/tape/ensemble_readers.py
dougbrn: Good catch, forgot to remove these from ensemble_readers.py as well!

src/tape/ensemble.py
dougbrn: Adding two conftest.py functions (one for from_parquet and one for from_dask_dataframes) as I think all the parameters for these loader functions live there

src/tape/ensemble.py
```diff
@@ -1284,19 +1286,32 @@ def save_ensemble(self, path=".", dirname="ensemble", additional_frames=True, **
...
 
+        # First look for an existing metadata.json file in the path
+        try:
+            with open(os.path.join(ens_path, "metadata.json"), "r") as oldfile:
+                # Reading from json file
+                old_metadata = json.load(oldfile)
+                old_subdirs = old_metadata["subdirs"]
+                # Delete any old subdirectories
+                for subdir in old_subdirs:
+                    shutil.rmtree(os.path.join(ens_path, subdir))
```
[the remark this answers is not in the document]
dougbrn: I'm thinking that whatever error shutils would provide would probably be as descriptive as anything we'd provide? Also not sure how easy it would be to test this, setting permissions mid-test seems messy...

src/tape/ensemble.py
```diff
@@ -1307,29 +1322,49 @@ def save_ensemble(self, path=".", dirname="ensemble", additional_frames=True, **
...
+        col_map = self.make_column_map()  # grab the current column_mapper
+
+        metadata = {
+            "subdirs": created_subdirs,
+            "known_divisions": divisions_known,
+            "column_mapper": col_map.map,
+        }
+        json_metadata = json.dumps(metadata, indent=4)
+
+        with open(os.path.join(ens_path, "metadata.json"), "w") as outfile:
```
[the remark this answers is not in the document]
dougbrn: Moved the parquet saving to the very end of the function in latest commit

wilsonbb (approved):
LGTM! :)

PR #353, opened 2024-01-19
Add Support for Sorting Lightcurves by Time

wilsonbb:
- [ ] My PR includes a link to the issue that I am addressing

Introduces `Ensemble.sort_lightcurves` which sorts lightcurves by their timestamp in ascending order. 

This PR closes https://github.com/lincc-frameworks/tape/issues/316 though we opt to provide a separate function for sorting lightcurves by their timestamp rather than adding additional functionality to the batch function. This also makes it easier for the user to only sort when needed rather than worrying about the arguments specified over repeated batch calls.

## Solution Description

Ideally we would use a multi-index of Object ID and timestamp for the Source table, but [Dask lacks support for this](https://github.com/dask/dask/issues/1493). Instead, we use the Object ID as the sorted index for both Source and Object tables, and a call to `Ensemble.sort_lightcurves` perform a per-partition Pandas `sort_values` for sorting the underlying Pandas dataframe by {Object ID, timestamp}  (Note that Dask `[sort_values](https://docs.dask.org/en/latest/generated/dask.dataframe.DataFrame.sort_values.html)` supports only sorting a single column). 

Because we aim for lightcurve cohesion, where the rows for each lightcurve are only on a single partition in the Source table, this per-partition sorting is all we need allowing us to escape some of the constraints of sorting entire Dask dataframes, especially allowing us to sort lightcurves on a lazy basis for only the partitions that we need. `Ensemble.sort_lightcurves` also has an optional parameter for whether to sort the lightcurves by band as well as time.

## Code Quality

## Project-Specific Pull Request Checklists
- [x] I have added a function that requires a sync_tables command, and have added the necessary sync_tables call

### New Feature Checklist
- [ ] I have updated the tutorial to highlight my new feature (if appropriate)
- [x] My change includes a breaking change
  - [x] My change includes backwards compatibility and deprecation warnings (if possible)


src/tape/ensemble.py
```diff
@@ -433,6 +433,28 @@ def check_lightcurve_cohesion(self):
                 return False
         return True
 
+    def sort_lightcurves(self):
```
hombit, line 436: Could we please make sorting key (everything going after `id_col`) to be an argument of this function? I can easily imagine a need to sort by `[id_col, band_col, time_col]` or something like that

hombit (commented):
Looks great!

What would actually happen if the condition of lightcurve cohesion is not met? Do we need a docstring telling something about that?

src/tape/ensemble.py
wilsonbb: It's definitely a good call to support sorting by band, but I decided not to add support for a fully general sorting key for now, and I decided to add a `by_band` parameter instead.

By just adding the `by_band` parameter, we can avoid users accidentally doing something like `[id_col, time_col, band_col]` which in general wouldn't be a sensible result. Of course users can still call `sort_values` themselves to sort arbitrary columns, but here we provide a simple helper function for the most common use cases.  Let me know your thoughts.

wilsonbb:
> Looks great!
> 
> What would actually happen if the condition of lightcurve cohesion is not met? Do we need a docstring telling something about that?

Added a note that this would fail to globally sort the table in the docstring.

src/tape/ensemble.py
hombit: Actually `[id_col, time_col, band_col]` also has some sense, for example for binned light curve. But I see your point, I'm fine with what we have now

hombit (approved):

PR #420, opened 2024-03-29
Add support for running batch on a single lightcurve

wilsonbb:
## Change Description
In https://github.com/lincc-frameworks/tape/issues/409 it was proposed to add a batch helper function `try_batch` which would allow `Ensemble.batch` to be tested on a single lightcurve. To accomplish that goal, here we instead add a new parameter `single_lc` to `Ensemble.batch`

When `single_lc` is provided an object ID, the function provided to `batch` is executed on that single lightcurve.

## Code Quality

### New Feature Checklist
- [x] I have updated the tutorial to highlight my new feature (if appropriate)

### DIFF
--- a/src/tape/ensemble.py
+++ b/src/tape/ensemble.py
@@ -1059,6 +1059,7 @@ class Ensemble:
         use_map=True,
         on=None,
         label="",
+        single_lc=None,
         **kwargs,
     ):
         """Run a function from tape.TimeSeries on the available ids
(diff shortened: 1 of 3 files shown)


src/tape/ensemble.py
```diff
@@ -1107,6 +1108,9 @@ def batch(
             source or object tables. If not specified, then the id column is
             used by default. For TAPE and `light-curve` functions this is
             populated automatically.
+        single_lc: `int`, optional
```
dougbrn, line 1111: I think it would be really helpful to also allow `single_lc=True`, where select_random_lightcurve is used to pick one for the user.

dougbrn (changes requested):
This looks good, just have one main change that I think should be made.

src/tape/ensemble.py
wilsonbb: Yeah that makes sense. To do that, I added an `id_only` parameter to `Ensemble.select_random_timeseries` so that it will only fetch the object id of a random lightcurve. But let me know if you would prefer that not added to the API and I can break that out differently.

dougbrn: Thanks! That looks like a good solution

dougbrn (approved):
Looks good!

PR #421, opened 2024-03-29
quiet the quickstart pip install

dougbrn:
I noticed that the pip install command in the quickstart guide has a lot of output (potentially because we now use the magic command?). To address this, I've added the quiet flag. Alternatively we could opt to return to using `!pip install lf-tape`


nevencaplar (approved):
I agree. I tried puting a semicolon at the end of juypter cell, but that does not stop terminal output being rendered.

PR #422, opened 2024-03-29
limit depth of toctrees

dougbrn:
Improves the table of contents view for each of our documentation subsections by just displaying the sections, rather than a full list of all subsections.


nevencaplar (approved):
So much better!

PR #433, opened 2024-04-24
Pin dask-expr to < 1.10.11 and revert "Add support for running batch on a single lightcurve"

wilsonbb:
Reverts lincc-frameworks/tape#420 and pins dask-expr <1.0.11 to address flakiness reported in https://github.com/lincc-frameworks/tape/issues/434. The plan is to do another PR for the batch single_lc changes in #420 after all of the batch-related flakiness has been addressed. 

Note that even with these changes we're still seeing the test_batch_by_band errors referenced in [lincc-frameworks/tape#434](https://app.slack.com/client/T06D204F2/1.0.11%20to%20address%20flakiness%20reported%20in%20%3Chttps://github.com/lincc-frameworks/tape/issues/434) consistently on the python 3.11.9 build, but this is still an improvement over the current test failures we've been seeing in https://github.com/lincc-frameworks/tape/issues/427

### DIFF
--- a/docs/tutorials/batch_showcase.ipynb
+++ b/docs/tutorials/batch_showcase.ipynb
@@ -124,43 +124,6 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "By default `Ensemble.batch` will apply your function across all light curves. However with the `single_lc` parameter you can test out your function on only a single object."
-   ]
-  },
-  {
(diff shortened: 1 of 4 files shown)


dougbrn (approved):
Thank you for looking into this. Did you try playing with the dask version itself at all? Their test suite is failing atm, and I wonder if pinned dask itself to something older would stabilize things

PR #438, opened 2024-04-26
add from_legacy_dataframe calls

### LINKED ISSUE #437: LSDB v0.2.1 Compatibility
LSDB v0.2.1 introduces support for dask versions >2024.3.0, but avoids using dask-expr in favor of using the lower level legacy interface (not via the config, but by importing the legacy classes directly). For TAPE, we need to modify our LSDB/hipscat bridge functions to handle this. The likely fix is to apply `from_legacy_dataframe` when loading LSDB catalog objects.

dougbrn:
Resolves #437


wilsonbb (approved):
LGTM, noting that `test_batch_by_band` failures pre-existed to this PR and are tracked here https://github.com/lincc-frameworks/tape/issues/434

PR #440, opened 2024-04-29
Reduced shared parquet file contention in testing

wilsonbb:
## Change Description
Since moving to dask-espr, we've seen increased test failures. In https://github.com/lincc-frameworks/tape/issues/434 there have been recent failures in `test_batch_by_band` where a flux column can sometimes be read in as NaNs during the `from_parquet` call used in that test invocation. Interestingly this only seems to occur when the tests are run as a suite and not when `test_batch_by_band` is run in isolation.

We can achieve the same effect by having the test build an ensemble using duplicate files of the source and object files. This appears to fix the test when run on github actions, but we should still investigate the underlying issue with `from_parquet` and then revert this change.

## Code Quality

### Bug Fix Checklist
- [ ] My fix includes a new test that breaks as a result of the bug (if possible)


dougbrn (approved):
Interesting that this works, makes me really want to have TAPE generate it's own test data via generator functions rather than use these test datasets...

